Issue #16, opened 2019-12-05
Integration does not load in 0.102.x


firstof9:
Integration does not load in 0.102.x, likely some behind the scene changes.

jseidl:
Same, I'm getting the following on the logs:

`Thu Dec 05 2019 14:47:41 GMT-0800 (Pacific Standard Time)
Error while setting up platform mail_and_packages
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 148, in _async_setup_platform
    task = async_create_setup_task()
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 104, in async_create_setup_task
    platform.setup_platform,
AttributeError: module 'custom_components.mail_and_packages.sensor' has no attribute 'setup_platform'
`

firstof9:
@jseidl This is due to having configuration entries in your config files, I'm working on allowing that as well as the integration configuration.

jseidl:
I've added the `mail_package.yaml` from `includes/packages/` file into my `packages` folder. Isn't supposed to add that?

firstof9:
No the instructions need adjusting as well, sorry, all you need is the custom_component directory.

jseidl:
Where should the configuration for the email server be entered?

firstof9:
> I've found the integration, took a while because the strings are not showing up for me.

This should be fixed in PR #17

PR #17, opened 2019-12-07
Fix config_flow

firstof9:
* Add remove functions to allow removal of sensors without requiring restart.
* Moved strings.json to .translations and renamed to en.json
* Simplified start up logging message


firstof9:
This should fix #16

firstof9:
@moralmunky I tested this locally on my production instance and a ha-dev docker instance, and it worked as long as my language was set to English in my HA profile.

You want to give it a whirl before merging?

firstof9:
I've been running this for the past 4 days and it's working well.

PR #27, opened 2020-02-12
Remove unit of time measurement

FuzzyMistborn:
Seems like a silly change but I don't believe the unit of measurement is required, and if you display the sensor as an entity in lovelace, by default it will say "Feb-12-2020 02:23 PM Time".  Removing the lines will make it read more correctly as just "Feb-12-2020 02:23 PM"


firstof9:
Whoops good catch.

Issue #1, opened 2019-06-02
Is my host correct.


codypet:
A couple of issues I seem to be having is I'm getting the sensors to come up, but I don't believe its logging in correctly.  I get notified that I have no packages even though I know there's a USPS package coming on June 4.  

My error looks like this:
`2019-06-02 04:22:37 ERROR (MainThread) [homeassistant.components.sensor] Error while setting up platform mail_and_packages
Traceback (most recent call last):
  File "/srv/homeassistant/lib/python3.5/site-packages/homeassistant/helpers/entity_platform.py", line 126, in _async_setup_platform
    SLOW_SETUP_MAX_WAIT, loop=hass.loop)
  File "/usr/lib/python3.5/asyncio/tasks.py", line 400, in wait_for
    return fut.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/mail_and_packages/sensor.py", line 85, in setup_platform
    add_devices([MailCheck(),USPS_Mail(),USPS_Delivering(),USPS_Delivered(),UPS_Delivering(),UPS_Delivered(),FEDEX_Delivering(),FEDEX_Delivered()])
  File "/home/homeassistant/.homeassistant/custom_components/mail_and_packages/sensor.py", line 137, in __init__
    self.update()
  File "/srv/homeassistant/lib/python3.5/site-packages/homeassistant/util/__init__.py", line 224, in wrapper
    result = method(*args, **kwargs)
  File "/home/homeassistant/.homeassistant/custom_components/mail_and_packages/sensor.py", line 159, in update
    account = login()
  File "/home/homeassistant/.homeassistant/custom_components/mail_and_packages/sensor.py", line 360, in login
    account = imaplib.IMAP4_SSL(host, port)
  File "/usr/lib/python3.5/imaplib.py", line 1269, in __init__
    IMAP4.__init__(self, host, port)
  File "/usr/lib/python3.5/imaplib.py", line 192, in __init__
    self._connect()
  File "/usr/lib/python3.5/imaplib.py", line 240, in _connect
    raise self.error(self.welcome)
imaplib.IMAP4.error: None`

[...]

PR #40, opened 2020-02-21
Efficiency patch

### LINKED ISSUE #5: -1 Packages in transit
In Transit shows -1 after deliveries are completed due to a new UPS subject line.

The new, additional, UPS subject line for a package out for delivery that you can view the status on a live map, "UPS Update: Follow Your Delivery on a Live Map".

### LINKED ISSUE #10: Slow start warnings
The component begins checking for email right away and appears to be causing a slowdown loading other components causing warnings.

### LINKED ISSUE #14: Each sensor logs into mailbox repeatedly

### LINKED ISSUE #32: Mail gif assumes shape of first mail piece
The mail I received today started with a square image and all subsequent rectangular images filled canvas area cutting off the left and right side. 

I had solved for this previously by resizing all the images proportionately within 700x315. I think this matched the no mail image. When assembling the gif the background was filled with white so you wouldn't see the outside parts of the pervious image if it was larger than the new image.

For reference:
`IMG_RESIZE_OPTIONS = ('convert -resize 700x315 ')
GIF_MAKER_OPTIONS = ('convert -delay 300 -loop 0 -coalesce -fill white '
                     '-dispose Background ')`

### LINKED ISSUE #33: GIF rotation is very slow
I find that 10 seconds is very slow rotation speed, especially if you have few mail pieces come through. It could be nice to add and option to adjust this speed when setting up the integration.

### LINKED ISSUE #34: Not picking up FedEx package as in transit
V0.07, installed via HACS using latest version of documentation
HomeAssistant version - 0.105.3

Got an email notification from FedEx from trackingupdates@fedex.com with title:

**FedEx Shipment 45280775XXXX: Your package is on its way** 
(Number redacted, but if you need it give me a way to DM you.)

But no state change on sensor.packages_in_transit or sensor.mail_fedex_packages.  

Link to email (address/package number redacted): https://drive.google.com/file/d/1foXa0mYWvv71c2C_vMgr256dzyXCt3YG/view?usp=sharing

Expected behavior:  Sensors would increase from 0 to 1 in at least packages in transit.

### LINKED ISSUE #36: In transit shows the same amount after deliveries recorded
Today I received 2 in delivering today messages and 2 delivered messages from USPS. I would expect the delivering count to reduce as the delivered count increased and the total USPS package count to only be 2 packages.

mailcard

firstof9:
* Only login to email once and pull all the emails needed in one swoop - Fixes #14 
* Changed GIF rotation time to 5 seconds - Fixes #33 
* Subtract delivered packages from 'in transit' packages, should no longer result in negative counts - Fixes #5 and Fixes #36 
* No more slow start warnings since we check the email in one call now - Fixes #10 
* Added aspect ratio calculations to keep image looking correctly - Fixes #32 
* Added additional subject searches for FedEx emails - Fixes #34 
* Bump version to 0.1.0 since we're no longer using Imagemagick and fixed the speed issues, seemed like a good milestone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ custom_components/mail_and_packages/.backup/__init__.py
 custom_components/mail_and_packages/.backup/camera.py
 custom_components/mail_and_packages/.backup/manifest.json
 custom_components/mail_and_packages/.backup/sensor.py
+notes.txt
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -1,5 +1,5 @@
 DOMAIN = "mail_and_packages"
-VERSION = "0.0.7"
+VERSION = "0.1.0"
 ISSUE_URL = "http://github.com/moralmunky/Home-Assistant-Mail-And-Packages"
 
 CONF_FOLDER = "folder"
@@ -28,6 +28,3 @@ FEDEX_Delivered_Subject = 'Your package has been delivered'
 
 GIF_FILE_NAME = 'mail_today.gif'
 GIF_DURATION = 10
-IMG_RESIZE_OPTIONS = ('convert -resize 700x315 ')
-GIF_MAKER_OPTIONS = ('convert -delay 300 -loop 0 -coalesce -fill white '
-                     '-dispose Background ')
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -28,4 +28,4 @@ FEDEX_Delivering_Subject = 'Delivery scheduled for today'
 FEDEX_Delivered_Subject = 'Your package has been delivered'
 
 GIF_FILE_NAME = 'mail_today.gif'
-GIF_DURATION = 10
+GIF_DURATION = 5
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -20,6 +20,7 @@ USPS_Delivered_Subject = 'Item Delivered'
 
 UPS_Email = 'mcinfo@ups.com'
 UPS_Delivering_Subject = 'UPS Update: Package Scheduled for Delivery Today'
+UPS_Delivering_Subject_2 = 'UPS Update: Follow Your Delivery on a Live Map'
 UPS_Delivered_Subject = 'Your UPS Package was delivered'
 
 FEDEX_Email = 'TrackingUpdates@fedex.com'
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -25,6 +25,7 @@ UPS_Delivered_Subject = 'Your UPS Package was delivered'
 
 FEDEX_Email = 'TrackingUpdates@fedex.com'
 FEDEX_Delivering_Subject = 'Delivery scheduled for today'
+FEDEX_Delivering_Subject_2 = 'Your package is scheduled for delivery today'
 FEDEX_Delivered_Subject = 'Your package has been delivered'
 
 GIF_FILE_NAME = 'mail_today.gif'
--- a/custom_components/mail_and_packages/manifest.json
+++ b/custom_components/mail_and_packages/manifest.json
@@ -5,5 +5,5 @@
   "dependencies": [],
   "codeowners": ["@moralmunky","@firstof9"],
   "config_flow": true,
-  "requirements": [ "imageio" ]
+  "requirements": [ "imageio", "scipy", "scikit-image" ]
 }
--- a/custom_components/mail_and_packages/sensor.py
+++ b/custom_components/mail_and_packages/sensor.py
@@ -9,6 +9,7 @@ Configuration code contribution from @firstof9 https://github.com/firstof9/
 import logging
 import asyncio
 import imageio as io
+from skimage.transform import resize
 import os
 import re
 import imaplib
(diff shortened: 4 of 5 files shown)


firstof9:
I've been running this for the past 3-4 days, other than the aspect ratio bit which I haven't been able to confirm yet, all else works great.

moralmunky:
FYI, I just add `today = '18-Feb-2020'`after line 901 in the sensor.py to pull a previous emails to test.

I had issues again with starting due to Mail Attachment.txt being added in the image array. I added the file name to line 828 in the sensor.py to the items to remove from the image array before processing.
`remove_terms = ['mailerProvidedImage', 'ra_0' , 'Mail Attachment.txt']` 

The image rotation continues to be an issue. I think the is a limitation with file camera. My observation is that there are only 10 seconds to get through all the mail images before it reloads the file. I don't think I ever encountered that before since I embedded the imagine the custom card instead of using the camera component.

So, my 4 mail images at 5 seconds each only gets through 2 images.
If I set the duration to 10, I only ever see 1 image. Not sure what the fix is here for a good viewing speed and being able to show the user all mail images.

For image resizing, doesn't seem to be there. I thought that maybe rescale is what should be used by I couldn't get the results. I think you would have to evaluate images widths and height because either could be out of the 317 x 700 perimeter. I'll keep looking into this.

firstof9:
> The image rotation continues to be an issue. I think the is a limitation with file camera. My observation is that there are only 10 seconds to get through all the mail images before it reloads the file. I don't think I ever encountered that before since I embedded the imagine the custom card instead of using the camera component.

Setting the lovelace card to "live" resolved this for me.

firstof9:
> For image resizing, doesn't seem to be there yet. I thought that maybe rescale is what should be used but I couldn't get the right results either. I think you would have to evaluate images widths and height because either could be out of the 317 x 700 parameter. I'll keep looking into this.

The resize function isn't working for you then? Or should I have it resize the width too?
```
# Resize images
# This should keep the aspect ratio of the images
#################################################


def resize_images(images):
    sized_images = []
    for image in images:
        if image.shape[1] < 700:
            wpercent = 700/image.shape[1]
            height = int(float(image.shape[0])*float(wpercent))
            sized_images.append(img_as_ubyte(resize(image, (height, 700))))
        else:
            sized_images.append(img_as_ubyte(resize(image, (317, 700))))

    return sized_images
```

moralmunky:
> Setting the lovelace card to "live" resolved this for me.

I see, you put in a picture entity card and that has the live setting. If this is the case we should probably make a note of the issue with a regular entity and change the 3 recommended configuration in the documentation to use the this card with the live settings on.

moralmunky:
Today I am getting the message that the image-no-mailpieces700.jpg doesn't exist inside the www path. I thought I read that the component now populates that folder? 

The image sizing is currently evaluating if the image width is less than 700. The intended function should be if the image width is greater than 700 do something. Even after scaling down the width, it is possible for the image height to still be too large for the intended canvas size. Also there is a type for the height, to match the no image for mail file the size 315px instead of 317.

**The logic**
From original image array, If image height is greater than 315
          get new width, make image 315 x new width
         add to first_pass_images array
else
         add unmodified image to first_pass_images array

From first_pass_images array, If image width is greater than 700
        get new height, make image new height x 700
       add to sized_images array
else
       add unmodified image to sized_images array

This is getting there. When stitching together in the gif, my first image was resized to w:660h:315 and that becomes the size of the gif. 

Two options I can think of now:
1. Find how skimage can add blank pixels to make all the images exactly 315 x 700 after they are sized to fit within that size.
2. Add a template image that is 315 x 700 to always be the first image of the array so the gif always takes on the right size. It could say 'mail today' or something so it's not just a blank slide.

`# Resize images
# This should keep the aspect ratio of the images
#################################################


def resize_images(images):
    first_pass_images = []
    sized_images = []
    for image in images:
        #1 height
        if image.shape[0] > 315:
            shape = "w:" + str(image.shape[1]) + "h:" + str(image.shape[0])
            wpercent = 315/image.shape[0]
            width = int(float(image.shape[1])*float(wpercent))
            size = "w:" + str(width) + "h:315"
            first_pass_images.append(img_as_ubyte(resize(image, (315, width))))
            _LOGGER.debug("Image size: %s", str(shape))
            _LOGGER.debug("Image too tall, new width: %s", str(width))
            _LOGGER.debug("resize percent: %s", str(wpercent))
            _LOGGER.debug("new size: %s", str(size))
        else:
            first_pass_images.append(image)
            _LOGGER.debug("Image fine")
            
    for image in first_pass_images:
        #1 check hegh
        if image.shape[1] > 700:
            shape = "w:" + str(image.shape[1]) + "h:" + str(image.shape[0])
            wpercent = 700/image.shape[1]
            height = int(float(image.shape[0])*float(wpercent))
            size = "w:700" + "h:" + str(height)
            sized_images.append(img_as_ubyte(resize(image, (height, 300))))
            _LOGGER.debug("Image size: %s", str(shape))
            _LOGGER.debug("Image too wide, new height: %s", str(height))
            _LOGGER.debug("resize percent: %s", str(wpercent))
            _LOGGER.debug("new size: %s", str(size))
        else:
            sized_images.append(image)
            _LOGGER.debug("Image fine")

    return sized_images`

moralmunky:
In my goosing I did find this, using a different library.
https://jdhao.github.io/2017/11/06/resize-image-to-square-with-padding/

moralmunky:
I'm going to merge this, add in the additional remove term, and add in the information for the camera card needing to be 'live' to the readme.

I think that would be good for a new release and the continue to work on the image sizing.

PR #327, opened 2021-01-08
Fix order sensitivity of sensor computation.

blueshiftlabs:
## Proposed change

The update() method of the sensor iterates through sensors in the order they appear in the \_resources list. Some sensors, such as the \*\_delivering and zpackages\_\* sensors, depend on the values of earlier sensors to compute their value. If the _resources list is not in exactly the right order, this causes these sensors to show the wrong value. Performing setup through the UI does not sort this list, and can cause the summary sensors (zpackages\_\*) to break.

To fix this issue, abstract the "fetch one sensor" logic into a fetch() method. That method can recursively call itself to fetch the value of any of its dependencies. The values are cached, so that a given sensor will only be computed once per update() pass, regardless of how many times that sensor is fetch()ed.

## Type of change

- [ ] Dependency upgrade
- [X] Bugfix (non-breaking change which fixes an issue)

### DIFF
--- a/custom_components/mail_and_packages/sensor.py
+++ b/custom_components/mail_and_packages/sensor.py
@@ -130,71 +130,76 @@ class EmailData:
 
             # Only update sensors we're intrested in
             for sensor in self._resources:
-                count = {}
-                if sensor == "usps_mail":
-                    count[sensor] = get_mails(
-                        account,
-                        self._img_out_path,
-                        self._gif_duration,
-                        self._image_name,
-                        self._generate_mp4,
-                    )
-                elif sensor == const.AMAZON_PACKAGES:
-                    count[sensor] = get_items(account, ATTR_COUNT, self._amazon_fwds)
-                    count[const.AMAZON_ORDER] = get_items(
-                        account, ATTR_ORDER, self._amazon_fwds
-                    )
-                elif sensor == const.AMAZON_HUB:
-                    value = amazon_hub(account, self._amazon_fwds)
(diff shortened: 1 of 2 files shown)


blueshiftlabs:
@firstof9 Looks like you're the one doing most of the reviews in this repo, please take a look.

firstof9:
Sorry for the late updates, this didn't pop into my notifications for some reason.
I'd like to test this against the 0.3.0 branch as well and see how the tests run over it before merging and release.

Thank you for the contribution.

custom_components/mail_and_packages/sensor.py
```diff
@@ -130,72 +130,77 @@ def update(self):
...
-                    count[sensor] = update_time()
-                else:
-                    count[sensor] = get_count(
-                        account, sensor, False, self._img_out_path, self._hass
-                    )[ATTR_COUNT]
-
-                data.update(count)
+                self.fetch(account, data, sensor)
+
+            self._data = data
```
firstof9, line 135: Updating `_data` twice

firstof9 (changes requested):
Please bump the version in const.py to 0.2.3 as well.

custom_components/mail_and_packages/sensor.py
```diff
@@ -130,72 +130,77 @@ def update(self):
...
             self._data = data
```
firstof9, line 136: This is the 2nd update

firstof9 (changes requested):

firstof9 (approved):

PR #393, opened 2021-03-04
Update version

jcgoette:


firstof9:
Versions are updated as part of our github workflow. We don't manually update them in the `const.py` or `mainfest.json` anymore.

jcgoette:
Per latest HA update, need in manifest.json. See Breaking Changes:

https://www.home-assistant.io/blog/2021/03/03/release-20213/#breaking-changes

firstof9:
If you are getting error messages about the version missing in Home Assistant, you can try out one of the beta releases, but the breaking changes listed, we've already taken care of.

PR #642, opened 2022-03-10
fix(shipper): increased detection for AusPost delivery emails

Luke-Beck:
## Proposed change

A lot of the email subjects (all in my case) from AusPost include the stores name in the middle of the subject, preventing the current search term from working.
"Your delivery is coming today" will only work if the store name isn't present.
In all the cases from my experience the subject will be in one of these 3 formats:
"Your delivery from Kmart is being prepared"
"Your delivery from Love Honey is on its way"
"Your delivery from Walter White is coming today" 

I have tested this change in my live Home Assistant install and it correctly finds my current deliveries.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [x] Update existing shipper

## Additional information

This change also finds AusPost deliveries that have just been assigned a tracking number and aren't necessarily out for delivery yet.  
It will find deliveries at every stage. Tracking assigned, In-Transit (between depots) and Delivering today.

### DIFF
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -194,7 +194,11 @@ SENSOR_DATA = {
     },
     "auspost_delivering": {
         "email": ["noreply@notifications.auspost.com.au"],
-        "subject": ["Your delivery is coming today"],
+        "subject": [
+            "is coming today",
+            "is on its way", 
+            "is being prepared"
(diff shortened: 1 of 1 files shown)


firstof9:
Thank you, but it looks like this was already corrected in the latest betas.

PR #758, opened 2022-11-28
feat(shippers): Add PostNL

AlexLamond:
## Breaking change

This release adds a small, possibly breaking, change - It adds that parcels can be in the state _exception, for when the parcel has either failed to deliver, has an issue and _processing is for when the process of being out for delivery has not yet begun

## Proposed change

To add PostNL as a carrier
To add the stages processing and exception

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [X] Adds a new shipper
- [ ] Update existing shipper

### DIFF
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -426,6 +426,29 @@ SENSOR_DATA = {
     },
     "dhl_parcel_nl_packages": {},
     "dhl_parcel_nl_tracking": {"pattern": ["[0-9A-Z]{12,24}"]},
+    # Post NL
+    "post_nl_processing": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket is ontvangen door PostNL"],
+    },
+    "post_nl_collection": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket ligt klaar"],
+    },
+    "post_nl_delivering": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket is onderweg", "De chauffer is onderweg"],
+    },
+    "post_nl_exception": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["We hebben je gemist"],
+    },
+    "post_nl_delivered": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket is bezorgd"],
+    },
+    "post_nl_packages": {},
+    "post_nl_tracking": {"pattern": ["3S?[0-9A-Z]{14}"]},
 }
 
 # Sensor definitions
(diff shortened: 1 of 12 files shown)


AlexLamond:
Looks like it's failing due to lacking test coverage, can someone advise me what needs added to get the test coverage bumped up? I see a number of carriers missing in tests/const.py

custom_components/mail_and_packages/helpers.py
```diff
@@ -381,8 +381,10 @@ def fetch(
         delivering = fetch(hass, config, account, data, f"{prefix}_delivering")
         delivered = fetch(hass, config, account, data, f"{prefix}_delivered")
         count[sensor] = delivering + delivered
-    elif "_delivering" in sensor:
+    elif "_delivering" in sensor or "_exception" in sensor or "_processing" in sensor:
         prefix = sensor.replace("_delivering", "")
+        prefix = sensor.replace("_exception", "")
+        prefix = sensor.replace("_processing", "")
```
firstof9, line 387: This change isn't needed exceptions are already handled in the function.

custom_components/mail_and_packages/const.py
```diff
@@ -733,6 +756,43 @@
...
+        native_unit_of_measurement="package(s)",
+        icon="mdi:dolly",
+        key="post_nl_preparing",
+    ),
+    "post_nl_collection": SensorEntityDescription(
+        name="Post NL Ready for Collection",
+        native_unit_of_measurement="package(s)",
+        icon="mdi:store-clock",
+        key="post_nl_ready_for_collection",
+    ),
```
firstof9, line 771: These are states we don't track.

custom_components/mail_and_packages/const.py
```diff
@@ -413,6 +413,29 @@
...
     "dhl_parcel_nl_tracking": {"pattern": ["[0-9A-Z]{12,24}"]},
+    # Post NL
+    "post_nl_processing": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket is ontvangen door PostNL"],
+    },
+    "post_nl_collection": {
+        "email": ["noreply@notificatie.postnl.nl"],
+        "subject": ["Je pakket ligt klaar"],
+    },
```
firstof9, line 424: These are states we don't track.

firstof9 (changes requested):

custom_components/mail_and_packages/const.py
AlexLamond: I think this would be nice to add however, since parcel shop collection can happen for failed parcels

custom_components/mail_and_packages/helpers.py
AlexLamond: I don't see this anywhere in the function. This compliments the two new states I added, processing and exception to be able to check which parcels are in those states. Exception state is important to know what parcels need attention

custom_components/mail_and_packages/const.py
firstof9: It's mostly out of the scope of the integration, we're only really interested in "heading to your home", "delivered to your home", "exceptions delivering your package" (if applicable), and the count of packages.

custom_components/mail_and_packages/helpers.py
firstof9: Exceptions are handled when they fall out of the bottom of the `elif` branches.

https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/dev/custom_components/mail_and_packages/helpers.py#L405-408

custom_components/mail_and_packages/const.py
AlexLamond: Still think it would be a nice addition, the integration comes across as parcel tracking, and this is a vital step in the process

firstof9: I'd start by remove the changes in `helpers.py` any added on sensors should fall out of the bottom of the `if` tree.

If you want to make sure the processing works correctly you'll need to supply a raw email, sanitized of any personal information, into the tests with the reset of the test emails.

Then you can use another test as an example of how to test your changes.
Example:

https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/dev/tests/test_helpers.py#L1069-L1074

and

This is where you'd read in the email for the above test:

https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/dev/tests/conftest.py#L452-L476

custom_components/mail_and_packages/const.py
```diff
@@ -733,6 +756,43 @@
...
+        name="Post NL Processing",
+        native_unit_of_measurement="package(s)",
+        icon="mdi:dolly",
+        key="post_nl_preparing",
+    ),
+    "post_nl_collection": SensorEntityDescription(
+        name="Post NL Ready for Collection",
+        native_unit_of_measurement="package(s)",
+        icon="mdi:store-clock",
+        key="post_nl_ready_for_collection",
```
firstof9, line 770:
```suggestion
        key="post_nl_collection",
```
They key needs to match the sensor.

firstof9 (changes requested):

vinnie1234:
Is there any new development for this request? PostNL would be great!

PR #763, opened 2022-12-03
fix(shipper): update DHL for Germany

CHirch:
Fixed German DHL sensor

## Breaking change

none

## Proposed change

German DHL sensor did not work. I adjusted the patterns so hopefully it will now detect mails correctly.

## Type of change

- [x] Update existing shipper


firstof9 (changes requested):
Please target the `dev` branch.

firstof9:
You'll need to make the edits off the `dev` branch rather than the `master` branch you created this from.

PR #818, opened 2023-04-19
feat(shipper): add Deutsche Post and DPD, updated shippers for german language and general fixes for shippers

Pr0mises:
## Proposed change
- Adding Deutsche Post Briefankündigung
- Adding DPD (kept dpd_com_pl, so no breaking change)
- Added/fixed german language to multiple providers
- Fixed german language

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [x] Adds a new shipper
- [x] Update existing shipper

## Additional information

I'm not sure if the hermes body excludes non german subjects or not. If yes feel free to change it.


custom_components/mail_and_packages/const.py
```diff
@@ -982,10 +1082,12 @@
...
     "inpost_pl",
     "dpd_com_pl",
+    "dpd",
     "gls",
     "dhl_parcel_nl",
     "bonshaw_distribution_network",
     "purolator",
     "intelcom",
     "post_nl",
+    "post_de",
```
firstof9, line 1092:
```suggestion
```

We only put shippers here that have both `delivering` and `delivered` sensors.

firstof9 (changes requested):

firstof9:
Looks like some tests will need updating on this.

custom_components/mail_and_packages/const.py
Pr0mises: I thought so, got a bit confused because of Poczta post, as it is included without delivered

Pr0mises:
Probably, I've no hass instance for checking against the tests in vscode rn (vm crashed). I only tested it on my production server

custom_components/mail_and_packages/const.py
```diff
@@ -275,11 +277,20 @@
...
+            "parcel is now with your local Hermes courier",
+            "Ihre Hermes Sendung",
+        ],
+        "body": [
+            "Voraussichtliche Zustellung",
+        ],
     },
     "hermes_packages": {},
-    "hermes_tracking": {"pattern": ["\\d{16}"]},
+    "hermes_tracking": {"pattern": ["\\d{11,16}"]},
```
firstof9, line 293: Are you sure the tracking number can be 11 or 16 digits?

Pr0mises: According to https://www.myhermes.de/empfangen/sendungsverfolgung/

Between 11 and 20 digits

firstof9:
Thanks!

PR #938, opened 2024-07-06
fix(shipper): add more buildinglink subject lines

raccettura:
Encountered a few more that I missed the first time around.  Mostly last-mile providers used by aliexpress and the like. The last one seems like a catch-all for anything not specifically in the system.


custom_components/mail_and_packages/const.py
```diff
@@ -594,7 +594,11 @@
             "Your USPS delivery has arrived",
```
firstof9, line 594: Can we try condensing lines 594-599 to:
```python
            "delivery has arrived",
```

firstof9 (changes requested):

custom_components/mail_and_packages/const.py
raccettura: Sounds reasonable to me.

firstof9 (approved):
Excellent, thanks!

PR #939, opened 2024-07-06
refactor: change the amazon config flow, use less searching

firstof9:
## Breaking change

Change the configuration for Amazon sensors, use less searches, has user setup amazon domain.

## Type of change

- [ ] Dependency upgrade
- [x] Breaking change (fix/feature causing existing functionality to break)
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR is related to issue: n/a


spitfire:
I live in Poland and while I generally use amazon.pl, I sometimes order stuff from other European amazon sites (using the same account). Is there a way to configure multiple entries?

firstof9:
> Is there a way to configure multiple entries?

Currently, you'd setup a second entry of the integration with another amazon domain. I'll see what I can do about using a list of domains so you only need a single configuration entry setup.

Issue #915, opened 2024-06-20
Add Sensor for delivered USPS mail.


ClaymoreTech4:
USPS added the ability to receive an email once mail was delivered.

USPS started new service to receive and email once mail is delivered,  below is a screenshot to set it up and the email you receive once you mail is delivered.

It would be fantastic to have a sensor to show Mail delivered to be able to run automations.

Thank you 

![Screenshot_20240620-105431.png](https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/assets/67112140/9d6ef06b-3ace-45da-a2c5-5b13aac7c64d)

￼

MAIL DELIVERY NOTIFICATION

￼

 MAIL

View Dashboard

￼

Your mail has been delivered today,
June 18!

Please note that Mail Delivery Notifications are not assurances that a specific mailpiece visible in your Informed Delivery Daily Digest Email or Dashboard has been delivered.

We value your feedback as we seek to continuously improve the accuracy of this feature.  After checking your mailbox today, please tell us about your experience with this notification:

￼￼￼ 

You subscribed to this feature with USPS® Product Technology Innovation, 475 L'Enfant Plaza SW, Washington, DC 20260. If you no longer wish to receive Mail Delivery Notifications, you can click here to unsubscribe.

If you need assistance, please email questions or concerns to USPSInformedDeliveryQuestions@usps.gov.

Copyright© 2024 United States Postal Service®. All Rights Reserved. The Eagle Logo is among the many trademarks of the U.S. Postal Service® . This is an automated email, please do not reply to this message. This message is for the designated recipient only and may contain privileged, proprietary, or otherwise private information. If you have received it in error, please delete. Any other use of the email by you is prohibited.

USPS.com | Privacy Policy | FAQs

firstof9:
What's the from address and the subject of the email?
This option isn't yet available in my area so I cannot test.

ClaymoreTech4:
![markup_1000006571.png](https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/assets/67112140/18bb0a61-cd14-4037-8a4a-14448c20d897)

firstof9:
Notes:

```python
    "usps_mail_delivered": {
        "email": [
            "USPSInformedDelivery@usps.gov",
            "USPSInformeddelivery@email.informeddelivery.usps.com",
            "USPSInformeddelivery@informeddelivery.usps.com",
            "USPS Informed Delivery",
        ],
[...]

PR #966, opened 2024-08-09
fix: fix get_resources not adding usps_mail_delivered to configuration selection

gpeavy:
## Proposed change

Fix bug in delivered USPS mail implementation.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

fetch() is never called with sensor="usps_mail_delivered" in process_emails() as resources does not contain the BinarySensors.
This is probably just a temporary fix.

- This PR is related to issue: #915


custom_components/mail_and_packages/helpers.py
```diff
@@ -217,6 +217,12 @@ def process_emails(hass: HomeAssistant, config: ConfigEntry) -> dict:
         except Exception as err:
             _LOGGER.error("Error updating sensor: %s reason: %s", sensor, err)
 
+    # Update usps_mail_delivered
+    try:
+        fetch(hass, config, account, data, "usps_mail_delivered")
```
firstof9, line 222: The sensor would already be in the `resources` and searched for twice.

gpeavy: I added some additional logging in my testing and I don't think it's in `resources`.  If `resources` is set by `get_resources()`, `SENSOR_TYPES` does not contain the BinarySensors. I think.

gpeavy: Please correct me in I am wrong.  This is my first time digging into HA. Does `config.get(CONF_RESOURCES)` get set by:
https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/4a397286698f3b3ea4d26542b279071b25716284/custom_components/mail_and_packages/config_flow.py#L211-L213
If so `get_resources()` here:
https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/4a397286698f3b3ea4d26542b279071b25716284/custom_components/mail_and_packages/helpers.py#L94-L103
uses `SENSOR_TYPES` defined here:
https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/4a397286698f3b3ea4d26542b279071b25716284/custom_components/mail_and_packages/const.py#L651
which does not include `usps_mail_delivered` defined here:
https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/4a397286698f3b3ea4d26542b279071b25716284/custom_components/mail_and_packages/const.py#L1157

firstof9: Ah you are correct, it's been excluded from the resources in error. That's where the bug needs to be fixed.

gpeavy: That was my guess. I wasn't exactly sure how to do that.

gpeavy: In addition to the Binary sensors, I think the Images sensors are missing too.

firstof9: I put in a PR on your fork that would solve the issue.

firstof9: > In addition to the Binary sensors, I think the Images sensors might be missing too, as in they are always created and not selected during setup.

This is by design (for now).

gpeavy: > I put in a PR on your fork that would solve the issue.

I'll pull it in and give it a try.

firstof9: Feel free to give it a run on a test HA instance if you like too, all the sensors will now be sorted alphabetically.

custom_components/mail_and_packages/const.py
```diff
@@ -4,10 +4,8 @@
 
 from typing import Final
 
-from homeassistant.components.binary_sensor import (
-    BinarySensorDeviceClass,
-    BinarySensorEntityDescription,
-)
+from .entity import MailandPackagesBinarySensorEntityDescription
```
firstof9, line 7:
```suggestion
from homeassistant.components.binary_sensor import BinarySensorDeviceClass
```

custom_components/mail_and_packages/const.py
```diff
@@ -4,10 +4,8 @@
...
+from homeassistant.components.binary_sensor import BinarySensorDeviceClass
```
firstof9, line 8:
```suggestion
from homeassistant.components.sensor import SensorDeviceClass, SensorEntityDescription
```

custom_components/mail_and_packages/const.py
```diff
@@ -4,10 +4,8 @@
...
 from homeassistant.components.sensor import SensorDeviceClass, SensorEntityDescription
```
firstof9, line 9:
```suggestion
from homeassistant.helpers.entity import EntityCategory
```

custom_components/mail_and_packages/const.py
```diff
@@ -4,10 +4,8 @@
...
 from homeassistant.helpers.entity import EntityCategory
```
firstof9, line 10:
```suggestion
```

custom_components/mail_and_packages/const.py
```diff
@@ -4,10 +4,8 @@
...
```
firstof9, line 11:
```suggestion
from .entity import MailandPackagesBinarySensorEntityDescription
```

firstof9 (changes requested):
This should fix the import linting errors.

custom_components/mail_and_packages/helpers.py
gpeavy: I've finally started to test it now.  Running into a weird situation where the Logbook and History show the sensor turning on, but the sensor still shows off. 
image

firstof9: if you refresh does it show "on"?
if so, this will be an bug in HA not updating the front end.

gpeavy: Interesting.  Yep. Issue with HA front end. I had clicked around to different views in HA and even restarted the Mail and Packages integration, but it wasn't until I refreshed the browser that it updated.

firstof9: I've noticed this issue after the 2024.8.2 update.

PR #1020, opened 2024-11-26
fix: update amazon france with french strings

BenJule:


firstof9:
The `amazon_search` function actually needs to be aligned with the `get_items` method for searching emails.

firstof9:
The issue you were trying to fix has been fixed in #1028
Thanks!

PR #1095, opened 2025-04-23
fix: adjust SSL contexts to use the non-cached versions provided by HA

StephenBrown2:
## Proposed change

Uses non-shared/cached SSL contexts when connecting via SSL or StartTLS. Fixes Fastmail connections.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Related Home Assistant Issue: https://github.com/home-assistant/core/issues/138573
Required Home Assistant change: https://github.com/home-assistant/core/pull/142653 (Assumed to be released with HA 2025.5.0)

- This PR fixes or closes issue: fixes #1052, closes #1092
- This PR is related to issue: https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/issues/1052

PR #1096, opened 2025-04-24
Use uv instead of pip

StephenBrown2:
## Proposed change

When bumping homeassistant to the dev version for testing a new function not yet in a released version, I found that pip could not resolve the dependencies when attempting to run the tests: https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/actions/runs/14628231584/job/41046210245?pr=1095 However, when I attempted to install them locally in a venv using `uv pip install -r requirements_test.txt`, I was able to get the package resolution to complete.

This changes to use uv with tox to manage dependencies, similar to Home Assistant, which has also switched to using uv: https://developers.home-assistant.io/blog/2024/04/03/build-images-with-uv/

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

I have made the minimal changes necessary to support using uv, however the rest of the config could be moved to `pyproject.toml`, and the linters could be consolidated to using [ruff](https://docs.astral.sh/ruff/faq/) (by the authors of uv), but I didn't want to disturb too much.

- This PR is related to PR: #1095

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -21,17 +21,14 @@ jobs:
     steps:
       - name: 📥 Checkout the repository
         uses: actions/checkout@v4
-      - name: 🛠️ Set up Python
-        uses: actions/setup-python@v5
         with:
-          fetch-depth: 2      
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+          fetch-depth: 2
+      - name: 🛠️ Set up uv with python ${{ matrix.python-version }}
+        uses: astral-sh/setup-uv@v5
         with:
           python-version: ${{ matrix.python-version }}
       - name: 📦 Install requirements
-        run: |
-          pip install tox tox-gh-actions          
+        run: uv tool install tox --with tox-uv --with tox-gh-actions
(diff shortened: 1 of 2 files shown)


StephenBrown2:
Force-pushed a potential fix for uv to only use it's managed versions, rather than the system python, which doesn't line up with the requested version.

StephenBrown2:
Well that is highly irregular. For some reason uv thinks that the latest version of homeassistant it can install for python 3.13 is v0.31.1, which is ancient, and thus requires an ancient version of pyyaml, which of course can no longer be built on the newer python:
```
help: `pyyaml` (v3.13) was included because `homeassistant` (v0.31.1)
        depends on `pyyaml`
```
But when I try locally with python 3.11, it resolves the latest version of homeassistant as 2024.3.3, and thus gets the later version of pyyaml.

Pinning homeassistant to the actual latest version does resolve everything correctly.

I've not run into an issue like this before, so I'll just leave this alone and just wait for the next beta to be released to properly test #1095.

StephenBrown2:
This likely requires more refactoring than is worth it at this time.

PR #1141, opened 2025-11-05
fix: update amazon delivery

ChrisLanks:
## Proposed change

There are a few updates/fixes for amazon in this PR.
1.  Amazon has changed their delivery email once again. I fixed the regex for this. 
2. I noticed a bug where checking for the amazon domain in the forwarded emails wasn't happening. So effectively a forwarded email that did not originally come from an amazon domain was being picked up. This resolves that issue.
3. We used to make the assumption that 1 order number was equal to 1 package. Unfortunately it seems this is no longer the case. Amazon had 1 order number for 2 packages, each had their own shipping data. So i updated the logic to no longer dedup the order numbers found. But instead, count how many emails there were for a given order number and check the dates on those specific email deliveries individually. This helped fix bugs in the overall "in transit" counter and "in delivery" counter.

I added a substantial amount of tests to test for all this. I have been running this for a few days and have not seen issues, though I have only been testing amazon deliveries.

## Type of change

- [ ] Dependency upgrade
- [X] Bugfix (non-breaking change which fixes an issue)
- [X] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Here is what I was seeing before when the logic was based on order number. The in transit deduped and showed 1, while delivered had 2 packages for 1 order number:
image

I then tested the logic by deleting the "delivered emails" (which puts the emails in a different folder to test the new logic in real time, and it accurately found the proper packages in transit).
image

I then restored the delivered package emails from amazon and got the correct delivered packages along with n transit decrementing properly:
image


custom_components/mail_and_packages/helpers.py
```diff
@@ -520,7 +522,7 @@ def fetch(
     amazon_fwds = cv.ensure_list_csv(config.get(CONF_AMAZON_FWDS))
     image_name = data[ATTR_IMAGE_NAME]
     amazon_image_name = data[ATTR_AMAZON_IMAGE]
-    amazon_days = config.get(CONF_AMAZON_DAYS)
+    amazon_days = int(config.get(CONF_AMAZON_DAYS, DEFAULT_AMAZON_DAYS))
```
firstof9:
```suggestion
    amazon_days = config.get(CONF_AMAZON_DAYS, DEFAULT_AMAZON_DAYS)
```

We shouldn't need to type this to `int` it should already be an `int` from the config.
If I am wrong, please correct me. :slightly_smiling_face:

ChrisLanks: I agree, it isn't necessarily necessary.  I can remove it. However, it would catch a case where someone may manually add : "3" in their config and this would allow it to parse correctly. 
Do you have a preference?

firstof9:
Looks good, thanks!

PR #1144, opened 2025-11-12
feat(fedex camera): Add fedex cam, refactor code

ChrisLanks:
## Breaking change

I renamed the walmart camera to have "delivered" in the name to match the other cameras

## Proposed change

I am adding a fedex camera for delivery. I noticed there is a picture when the image is dropped off so I added it as a camera.

I also refactored the code. I combined the UPS, Walmart and FedEx logic into a single function. I hope this cleans up the duplicate logic and enables this to scale better. There is now a single image extractor function and the count() function does the majority of the work.

I'm interested in what you think of this approach compared to duplicating code everywhere. If we would rather duplicate code, that's fine and an option  But given all the feature requests, I wanted to get closer to adding a single function/config spot to add a new camera.  This allows a user to add it to const.py and it should automatically get extracted and added to "generic_camera" now. 
However, if they want an individual camera for the shipper or to customize the no_delivery jpeg, they still need to update config_flow, and majority of files (all languages files, etc) and add a lot of tests. 

I also think its possible if someone doesnt go from beta to beta, they may miss some config_flow upgrades. So I added some extra steps in the latest migration to account for that. 

The "breaking change" is I renamed the walmart camera from "walmart camera" to "walmart delivered camera" to start to standardize the names.

I also added a todays_date() function everything calls for the todays date. It enables testing in the past easier. You can set the date to the day before or some random date to test against past emails. Really helpful. 

I also discovered a bug where images may not be the right direction in the generic camera when the gif is created. Before it just took the images and made a gif and ignored the exif metadata. It now will orientate the image the right way.

I also know the tests are broken, but I need to (or if anyone has time) need to get back and fix them.  I am running this live now and testing in real time. 

## Type of change

- [ ] Dependency upgrade
- [X] Breaking change (fix/feature causing existing functionality to break)
- [ ] Documentation update
- [ ] Adds a new shipper
- [X] Update existing shipper

## Additional information

This is both the camera for generic and fedex:

image

And when in transit:

image


firstof9:
I am all for de-duplication of code.

firstof9:
Looks like this will need a rebase.

ChrisLanks:
Rebasing has caused so many conflicts and seems to not be obvious to me whats changing and wear when there arent conflicts. I was working through them a few weeks ago and got caught up in the holidays. I may need to instead go through this new forward logic and add that functionality to my code vs rebasing.
For example, i noticed the function to set the date and time in 1 spot disappeared. 

Unless you have a better idea?

firstof9:
Is there a way to just do the fedex cam part on it's own?

ChrisLanks:
So i took my original PR, and then I went through the forward emails code and added it back in. I will need to retest this with actual packages. But I believe this has all the updates.

I have been running my code (without the forwarded emails) for several weeks now.

ChrisLanks:
image
I can confirm after adding these changes, i am getting what I expect in terms of packages (4 from amazon, 3 from usps as the shipper)

ChrisLanks:
rebased

firstof9:
Give me a bit to review, I'll see if I can sneak in some tests to get the coverage higher :smile:

firstof9:
I can tighten up the tests later, have you been running this latest code base @ChrisLanks ?

ChrisLanks:
@firstof9 I have been, but since i rebased yesterday I have only had like half a day to run things and no new packages. Let me play with it today. I'll update the time to a few days ago now that its centralized on that 1 function and see if it still behaves as I expect.

ChrisLanks:
So far, this makes sense (today):

image


Dec 5th:  - makes sense (2 pieces of mail, 4 packages)

image

image

Still working when i delete the delivered notifications. It shows in transit correctly.

ChrisLanks:
I know this is closed and merged, but wanted to post that its still working with fedex today and I will update when i get the photo once delivered
image

ChrisLanks:
image
Complete!

PR #1235, opened 2026-04-20
fix: normalize forwarded_emails string to list before concatenation in shippers

virtitnerd:
## Proposed change

`CONF_FORWARDED_EMAILS` was being stored as a raw comma-separated string in the config entry (e.g. `"me@example.com, other@example.com"`), but both `USPSShipper` and `GenericShipper` attempted to concatenate it directly with a list of carrier email addresses, producing a `TypeError: can only concatenate str (not "list") to str` on every update cycle for users with forwarded email addresses configured.

Three changes are included:

- **`shippers/usps.py`** — normalize `forwarded_emails` from str to list on read before concatenation
- **`shippers/generic.py`** — same normalization
- **`config_flow.py`** — convert the validated string to a list on save, consistent with how `CONF_AMAZON_FWDS` is handled, so newly saved/reconfigured entries store the correct type going forward

The shipper-side fix is defensive and handles all existing config entries immediately without requiring reconfiguration.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: n/a
- This PR is related to issue: n/a

**Error seen in logs prior to fix:**
```
ERROR (MainThread) [custom_components.mail_and_packages.coordinator] Error processing shipper generic: can only concatenate str (not "list") to str
ERROR (MainThread) [custom_components.mail_and_packages.coordinator] Error processing shipper usps: can only concatenate str (not "list") to str
```

Issue #1214, opened 2026-03-27
ISSUE: Stuck in initializing / timing out since 0.5.0


rscheuermann:
<img width="602" height="286" alt="Image" src="https://github.com/user-attachments/assets/ba02fe8b-d663-46c5-9a17-4584f0fce0d4" />

**Describe the bug**
Mail and Packages is stuck in 'initializing' since upgrading to 0.5.0. I tried recreating the entity config, and it connects to IMAP fine, but just sits in initializing for a long time then eventually fails setup. No entities are created. It seems to be timing out after 5 mins. It never did this before. 5 minutes is a long time. This has to be an implementation issue.

**Environment (please complete the following information):**
 - Type of device: Synology
 - Home Assistant version: 2026.3.4
 - Component version: 0.5.0

**Logs**
Paste your error logs here.

2026-03-27 12:15:24.753 ERROR (MainThread) [custom_components.mail_and_packages] Timeout fetching Mail and Packages (imap.mail.me.com) data
2026-03-27 12:15:24.753 DEBUG (MainThread) [custom_components.mail_and_packages] Full error:
Traceback (most recent call last):
  File "/config/custom_components/mail_and_packages/__init__.py", line 303, in _async_update_data
    data = await process_emails(self.hass, self.config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/mail_and_packages/helpers.py", line 328, in process_emails
    await fetch(hass, config, account, data, sensor)
  File "/config/custom_components/mail_and_packages/helpers.py", line 701, in fetch
    await get_count(
    ...<9 lines>...
    )
  File "/config/custom_components/mail_and_packages/helpers.py", line 1571, in get_count
    amazon_mentions = await find_text(
                      ^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "/config/custom_components/mail_and_packages/helpers.py", line 1807, in find_text
    matches, value = await _scan_email_for_text(account, i, patterns, body_count)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/mail_and_packages/helpers.py", line 1767, in _scan_email_for_text
    data = (await email_fetch(account, email_id, "(RFC822)"))[1]
[...]

Issue #893, opened 2024-05-23
ISSUE: USPS Informed Delivery digest new layout


tjhorner:
**Describe the bug**

The USPS Informed Delivery digest has a new email subject and HTML layout starting on May 23, so Mail and Packages no longer detects it. Just posting an issue here as a heads up!

**Environment (please complete the following information):**

 - OS: HAOS
 - Type of device: NUC
 - Home Assistant version: 2024.5.2
 - Hassio/Docker/Core? Hassio
 - Component version: 0.3.25

**Logs**

n/a

**Screenshots**

n/a

**Additional context**

I can provide a raw `.eml` file upon request to assist with implementation. I'm not uploading it here publicly due to the risk of leaking PII.

firstof9:
Technically we don't parse the HTML version of the email, we parse the plain text version.

tjhorner:
Ah you're right, didn't notice that. It also seems the changed subject line should not affect it. After reviewing the source some more it appears the changes USPS made shouldn't affect this integration. Sorry for the noise, then! I'll close this issue now.

firstof9:
> Sorry for the noise

Perfectly fine, you were keeping an eye out and everyone appreciates the vigilance.

PR #1237, opened 2026-04-22
fix: camera reload, USPS placeholder detection, IMAP leaks, and forwarded_emails migration

### LINKED ISSUE #897: ISSUE: Detected blocking call to open inside the event loop (HA 2024.6 beta)
**Describe the bug**
Updated to Home Assistant 2024.6 beta and there's warnings in logs related to Mail and Packages custom integration. Says to report to integration author, so here goes.

Home Assistant OS / core version 2024.6.0b3 / Mail and Packages version 0.3.25

**Logs**
```
Logger: homeassistant.util.loop
Source: util/loop.py:81
First occurred: May 30, 2024 at 11:29:09 PM (64 occurrences)
Last logged: 1:54:25 AM

Detected blocking call to open inside the event loop by custom integration 'mail_and_packages' at custom_components/mail_and_packages/helpers.py, line 314: with open(filename, "rb") as file: (offender: /config/custom_components/mail_and_packages/helpers.py, line 314: with open(filename, "rb") as file:), please create a bug report at https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/issues
```

### LINKED ISSUE #1004: ISSUE: Could not read camera Mail Amazon Delivery Camera image from file
**Describe the bug**
I'm getting this error of being unable to read the Amazon picture, even if nothing is being delivered today.  It tends to happen hundreds of times a day.

 - OS: [e.g. HassOS
 - Type of device: Unraid VM
 - Home Assistant version: 2024.11.2
 - Component version: 0.4.1

**Logs**
This error originated from a custom integration.

Logger: custom_components.mail_and_packages.camera
Source: custom_components/mail_and_packages/camera.py:128
integration: Mail and Packages (documentation, issues)
First occurred: 12:46:10 PM (1 occurrences)
Last logged: 12:46:10 PM

Could not read camera Mail Amazon Delivery Camera image from file: /config/custom_components/mail_and_packages/images/amazon/9c19849c-7386-4a01-8e96-608099fad1df.jpg

virtitnerd:
## Proposed change

This PR addresses several reliability, security, and compatibility issues discovered during a review of the integration's codebase.

**Security / reliability fixes:**
- Mask OAuth token from error logs to prevent credential leakage
- Guarantee IMAP logout on folder selection failure (previously the connection could be left open)
- Add timeout and size limit to Amazon image downloads to prevent hung requests or memory exhaustion
- Close IMAP connections in config flow `_get_mailboxes` and `_validate_login` after use
- Close image file handle after read in `async_camera_image`
- Replace `subprocess.call` with `subprocess.run` in `generate_grid_img` and handle errors

**Config migration (version 17 → 18):**
- Migrate `CONF_FORWARDED_EMAILS` from a bare string to a list, matching how all other multi-value config options are stored
- Add `CONF_FORWARDED_EMAILS` to diagnostics redaction (it was previously exposed in plain text)

**Camera update reliability:**
- Remove polling (`should_poll = False`) from all camera entities in favour of the existing coordinator push model — cameras were redundantly polling every 30 seconds on top of coordinator-driven updates
- Fix cameras stuck on the default "no mail" image after integration reload: the HA camera proxy cached the default image on first state write before `update_file_path()` had run; cameras now resolve the correct file path in `async_added_to_hass` before any state is written
- Replace synchronous `open()` calls in camera with `Path.open()` to eliminate blocking I/O warnings in HA 2024.6+

**USPS Informed Delivery compatibility:**
- USPS changed their email format — the old `image-no-mailpieces700.jpg` filename reference no longer appears in emails; unscanned mailpieces are now indicated by a `<div id="mailpiece-with-no-image-id">` block
- Detection is performed on the properly QP-decoded HTML inside `_extract_usps_images` rather than on the raw RFC822 wire bytes, eliminating the risk of a quoted-printable soft line break silently breaking the match
- The legacy `image-no-mailpieces700.jpg` plain-text reference check is retained for older email formats

**Code quality:**
- Deduplicate `CONF_AMAZON_FWDS` assignment in `_validate_user_input`
- Replace `issubset` guards and direct dict access with `.get()` throughout coordinator/sensor code
- Annotate intentional broad exception catches with `# noqa: BLE001`
- Add debug log to `_update_usps_camera` for parity with all other camera types
- Tighten `.gitignore`: add explicit `__pycache__/` rule and collapse per-shipper image directory rules into one comprehensive rule

**Test suite:**
- Add `resolve_entity_id()` helper to `conftest.py` that resolves entity IDs by stable `unique_id` substring rather than by the generated entity ID string — HA 2026.1.x (CI) and HA 2026.4.x (local) differ in whether the device name is prepended, so hardcoded entity ID strings broke across versions
- Change all camera `friendly_name` assertions from exact equality (`==`) to substring checks (`in`) for the same cross-version compatibility reason
- Add 9 new tests to reach 100% patch coverage: `_migrate_version_17` string-to-list and `"(none)"` removal paths; `selectfolder` exception triggering IMAP logout + `UpdateFailed`; `async_camera_image` cache-hit fast path; `generate_grid_img` FFmpeg `CalledProcessError` handling; and four `download_amazon_img` early-return branches (non-200 status, non-image content-type, content-length over limit, downloaded data over limit)
- Fix `test_device_diagnostics` isolation flake caused by `mock_update.return_value` pointing to a shared mutable dict that `_binary_sensor_update` mutated across tests; changed to `side_effect` returning a fresh copy each call
- Add test for new USPS no-image placeholder format

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [x] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #897, fixes #1004, fixes #1164
- This PR is related to issue: #893, #1214, #1232, #1235

> **Note on config migration:** Users upgrading from a version where `CONF_FORWARDED_EMAILS` was stored as a string will be automatically migrated to the list format on first load (version 17 → 18). No manual action required.


virtitnerd:
Not sure why it's failing on the CI test but not locally

Edit: Looks like CI is using an older version of homeassistant even though there is no target version in requirements_test.txt.

It's seems to be having issues with the entity naming prefix between the CI version and local (2026.4.3)

custom_components/mail_and_packages/utils/amazon.py
```diff
@@ -39,6 +39,8 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+_MAX_IMAGE_SIZE = 10 * 1024 * 1024  # 10 MB
```
firstof9, line 42: Do we have a reason to limit the file size here?

firstof9:
> Looks like CI is using an older version of homeassistant even though there is no target version in requirements_test.txt.

Yes, python 3.13 test should be removed soon, since the latest version of home assistant use python 3.14 now.

custom_components/mail_and_packages/utils/amazon.py
virtitnerd: The limit guards against a malformed or unusually large image URL causing resp.read() to buffer an arbitrarily large payload into memory before the executor write — particularly relevant on memory-constrained HA hardware like a Raspberry Pi. The total=30 timeout is complementary but doesn't bound size; on a fast connection you could still receive hundreds of MB inside 30 seconds. Real delivery photos are consistently under 2 MB, so 10 MB is a safe cap with no realistic false positives.

We could increase it from 10MB if we feel it's possible a larger image would come in, But I'd be expecting Amazon to also compress all of the images being used to save on storage

firstof9: Sounds reasonable.

firstof9:
Thanks!

PR #1267, opened 2026-06-13
fix: surface scan time-budget overruns and clarify the timeout setting

jasonjhofmann:
## Problem

`MailDataUpdateCoordinator._async_update_data` wraps the *entire* update cycle in `async with asyncio.timeout(self.timeout):`. The configured value (`CONF_IMAP_TIMEOUT`, default 60s) is therefore the budget for login **plus** every sequential per-carrier IMAP search — not just the connection. On a large mailbox this fires, and the user sees an opaque `Timeout fetching ... data` with no indication of how long it ran, what the budget was, or which carrier was mid-search. The config field is also mislabeled "Time in seconds before connection timeout", reinforcing the wrong mental model.

## Fix (surgical)

- **coordinator:** catch `TimeoutError` at the `asyncio.timeout` boundary, log an actionable ERROR (elapsed vs. budget + remediation: raise the limit, use a dedicated folder, shrink the days-back window, or enable fewer carriers) and re-raise unchanged so HA still marks the update failed. A monotonic clock is captured before the block. The existing inner `try/except`/`return` is untouched; `CancelledError` (BaseException) raised inside by the timeout is not swallowed by the inner `except Exception`, so only a genuine budget overrun reaches the new handler.
- **coordinator:** add per-shipper DEBUG timing so the slow carrier is identifiable in debug logs.
- **`__init__`:** when `last_exception` stringifies empty (a bare `TimeoutError`), fall back to the exception type name so the setup error log isn't blank.
- **strings/translations:** relabel the field to *"Mailbox scan time limit — total seconds for the entire scan, not just connecting (minimum 10)"*.
- **tests:** new focused test asserts the timeout path logs the actionable error and re-raises.

No behavior change on the success path; the timeout still propagates and the update is still marked failed. `ruff` clean; full test suite passes.


secondof9:
## Code Review Summary

### 🔴 Critical
- **`coordinator.py:113-158`** — Wrapped the entire `_async_update_data` method in `asyncio.timeout` and caught `TimeoutError` at the boundary. This is a **correct but potentially confusing design** because the timeout budget includes login plus every sequential carrier search, not just the connection. The error message accurately reflects this, but users will still see an opaque failure. No actual critical bug—just a UX/log clarity issue.
- **`__init__.py:143-146`** — Added a fallback to `type(exc).__name__` when `str(exc)` is empty (e.g., bare `TimeoutError`). Good defensive detail, though a bare `TimeoutError` should still surface the exception type itself.

### ⚠️ Warnings
- **`coordinator.py:113`** — Capturing `monotonic()` before the timeout block is excellent for actionable logs. No warning needed.
- **`coordinator.py:308-314`** — Added `shipper_start` timing with a `finally` block to log per-carrier duration. Excellent for identifying slow carriers in debug logs.
- **`strings.json` & `translations/en.json`** — Relabeled `imap_timeout` to clarify it covers the **entire scan**, not just connecting. The new wording is verbose but accurate. Consider whether this belongs in a tooltip/field description instead of a static string, but that's an UI/UX trade-off, not a code warning.

### 💡 Suggestions
- **`coordinator.py:113-158`** — Consider moving `asyncio.timeout` **inside** `process_emails()` (or wrapping the individual carrier calls) so the outer catch block sees only `UpdateFailed` on OAuth/token errors, not `TimeoutError`. This would keep the inner `try/except` focused on its original domain and avoid the "bare `TimeoutError`" fallback pattern. The current approach is simpler but mixes concerns.
- **`coordinator.py:308-314`** — The `finally` block logs a debug message. If the shipper fails, this will still fire and could be misleading ("processed in Xs" when it actually errored). Consider making the timing log conditional on success, or moving it to the `except` block after a re-raise decision.

### ✅ Looks Good
- **`tests/test_coordinator.py:106-131`** — New test `test_async_update_data_timeout_logs_and_reraises` explicitly asserts the timeout path logs an actionable error and re-raises. This directly validates the new behavior and prevents regression.
- **`__init__.py:143-146`** — The fallback to `type(exc).__name__` is a minimal, targeted improvement for empty error strings. No over-engineering.

---

firstof9 (changes requested):
Please review PR comments.

jasonjhofmann:
Thanks for the thorough review. Pushed `cfe3f79`. Walking each item:

**🔴 Critical 1 — timeout wraps the whole scan, opaque failure**
Adopted in spirit, no code change needed. The timeout's *scope* is unchanged from before this PR — `_async_update_data` has always wrapped the entire update in `asyncio.timeout(self.timeout)`; login + every carrier search were already inside the budget. The whole point of this PR is to make the resulting failure **non-opaque**: the new `except TimeoutError` logs the budget, the elapsed time, and exactly which knobs to turn. Combined with the relabeled setting, the "opaque failure" is what's being fixed, so I'd keep the wrapping where it is.

**🔴 Critical 2 — bare `TimeoutError` should surface the exception type**
Already covered. In `__init__.py`, `detail = (str(exc) or type(exc).__name__) if exc else "unknown error"` — `str(TimeoutError())` is `""`, so it falls through to `type(exc).__name__` → `"TimeoutError"`. So a bare timeout does surface its type rather than logging an empty string.

**⚠️ Warning — the relabeled string is verbose; consider a field description**
Adopted. Shortened the label back to `Mailbox scan time limit (seconds, minimum 10)` and moved the "covers the entire scan — login plus every per-carrier search — not just connecting" clarification into `data_description.imap_timeout` in both the config and reconfigure step-2 flows (+ `translations/en.json`). HA renders that as helper text under the field, which is the idiomatic home for it.

**💡 Suggestion 1 — move `asyncio.timeout` inside `process_emails()`**
Pushing back. The OAuth token refresh runs *before* `process_emails()`, inside the current timeout; moving the deadline into `process_emails` would leave token refresh unbounded and weaken the coordinator's "the whole update is bounded" contract. The separate `except TimeoutError` isn't really mixed concerns — the inner `try/except` maps domain errors to `UpdateFailed`, while the outer handler exists specifically because the deadline is the one failure mode that needs elapsed-time context in its log. I'd rather keep one end-to-end budget than split it.

**💡 Suggestion 2 — the `finally` timing log fires on failure too ("processed in Xs")**
Adopted (`cfe3f79`). Added a `success` flag; the debug line now reads `Shipper X processed in Ys` on success and `Shipper X failed in Ys` when the shipper raised, so the timing is never misleading.

Full suite green (517 passed, 100% coverage) on py314.

firstof9 (approved):

firstof9:
Thanks!

PR #1268, opened 2026-06-13
fix: camera falls back to bundled placeholder instead of HTTP 500 when image file is missing

jasonjhofmann:
## Problem

`MailCam.async_camera_image` reads `self._file_path` and, on `FileNotFoundError`, logged a debug line and returned `None`. Home Assistant's camera proxy turns a `None` image response into **HTTP 500**, so the dashboard renders a broken tile.

This happens whenever a camera's `_file_path` points at a file that isn't on disk:
- the generic camera's `generic_deliveries.gif` before any generic delivery has ever been written;
- any per-shipper delivery image that was cleaned up or not yet generated.

## Fix

Every shipper already ships a bundled placeholder in the package (`no_deliveries_<base>.jpg`, `no_deliveries_generic.jpg`, or `mail_none.gif` for USPS). The camera now falls back to that placeholder instead of returning `None`, so it never 500s:

- `__init__` captures the bundled placeholder once as `self._default_image_path`, so it's always available regardless of later `_file_path` changes.
- On `FileNotFoundError` for the primary path, `async_camera_image` reads the placeholder, caches against it, and returns its bytes. It only returns `None` if the shipped placeholder is *also* unreadable (should never happen) — logged at WARNING.

The cache fast-path and successful-read caching are preserved; `update_file_path()` still invalidates the cache, so a later real delivery image refreshes correctly.

## Tests

- New `test_async_camera_image_falls_back_to_placeholder` asserts placeholder bytes are returned (not `None`) and the cache is keyed to the placeholder path.
- `test_async_camera_image_file_error` now covers the both-missing case (returns `None` + WARNING).
- Two pre-existing tests that asserted the old `None` behavior were updated to the new fallback contract (the old `None` return was the bug).

Full suite passes; `ruff` clean.

### DIFF
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -145,6 +145,12 @@ class MailCam(CoordinatorEntity, Camera):
         else:
             self._file_path = f"{Path(__file__).parent}/{default_image}"
 
+        # Canonical bundled placeholder for this camera. Always available
+        # regardless of later _file_path changes, so a missing delivery image
+        # can fall back to it instead of returning None (which makes the HA
+        # camera proxy serve HTTP 500).
+        self._default_image_path = f"{Path(__file__).parent}/{default_image}"
+
         self._cached_image_path: str | None = None
         self._cached_image_bytes: bytes | None = None
         self._last_delivery_images: list[str] | None = None
@@ -178,10 +184,34 @@ class MailCam(CoordinatorEntity, Camera):
             )
         except FileNotFoundError:
             _LOGGER.debug(
-                "Could not read camera %s image from file: %s",
+                "Could not read camera %s image from file: %s; "
+                "falling back to bundled placeholder %s",
                 self._name,
                 self._file_path,
+                self._default_image_path,
             )
+            # Fall back to the bundled placeholder so the camera proxy never
+            # serves HTTP 500 when a delivery image is missing on disk.
+            if self._default_image_path and self._file_path != self._default_image_path:
+                try:
+                    image_bytes = await self.hass.async_add_executor_job(
+                        _read_file, self._default_image_path
+                    )
+                except FileNotFoundError:
+                    _LOGGER.warning(
+                        "Camera %s placeholder image also missing: %s",
+                        self._name,
+                        self._default_image_path,
+                    )
+                    return None
+                else:
+                    # Cache against the placeholder path so repeated reads are
+                    # cheap; update_file_path() invalidates the cache when the
+                    # primary path changes.
+                    self._cached_image_path = self._default_image_path
+                    self._cached_image_bytes = image_bytes
+                    return image_bytes
+            return None
         else:
             self._cached_image_path = self._file_path
             self._cached_image_bytes = image_bytes
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -191,8 +191,12 @@ class MailCam(CoordinatorEntity, Camera):
                 self._default_image_path,
             )
             # Fall back to the bundled placeholder so the camera proxy never
-            # serves HTTP 500 when a delivery image is missing on disk.
-            if self._default_image_path and self._file_path != self._default_image_path:
+            # serves HTTP 500 when a delivery image is missing on disk. Skip the
+            # fallback when the primary path already IS the placeholder (true for
+            # non-custom cameras, where _file_path and _default_image_path are
+            # both built from the same default image): re-reading the same
+            # missing file would only fail again, so return None instead.
+            if self._file_path != self._default_image_path:
                 try:
                     image_bytes = await self.hass.async_add_executor_job(
                         _read_file, self._default_image_path
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -284,7 +284,7 @@ async def test_async_camera_image_file_error(
     mock_update,
     caplog,
 ):
-    """Test async_camera_image function."""
+    """Test async_camera_image returns None when both image and placeholder are missing."""
     with (
         patch("os.path.exists", return_value=True),
         patch("os.access", return_value=False),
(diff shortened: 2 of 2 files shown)


secondof9:
## Code Review — PR #1268

### 🔴 Critical
**File:** `custom_components/mail_and_packages/camera.py`  
**Line:** 195-207

- **Failure Mechanic:** The guard `if self._default_image_path and self._file_path != self._default_image_path:` is overly defensive. The comparison is redundant — the constructor explicitly sets `_default_image_path` only when `default_image` is non-empty, so if the constructor path is entered, both are non-empty. The `and` clause will always be true.
- **Suggestion:** Strip the redundant comparison and simplify the guard:
  ```python
  if self._default_image_path:
  ```
  The explicit `and self._file_path != self._default_image_path` offers no functional benefit and adds cognitive overhead.

### ⚠️ Warnings
**File:** `tests/test_camera.py`  
**Line:** 311-317

- **Failure Mechanic:** The test patches `os.path.exists` to return `True`, but then patches `Path(...).open` to raise `FileNotFoundError`. This is a contradictory setup — the test claims to verify file-not-found behavior while asserting existence. The test author likely intended the `exists` patch to represent a race condition, but it's not documented and could mislead future reviewers.
- **Suggestion:** Remove the `os.path.exists` patch or add a comment explaining the race condition it's simulating. The test's primary goal is to verify fallback to `_default_image_path`, and the `open` side-effect simulation already covers that.

### ✅ Looks Good
**File:** `custom_components/mail_and_packages/camera.py`  
**Lines:** 178-212

- The new fallback logic is well-crafted: it catches `FileNotFoundError`, falls back to `_default_image_path`, and only returns `None` if the bundled placeholder is also unreadable (which should never happen — logged at WARNING). The caching behavior is keyed to the placeholder path, so repeated reads are cheap; `update_file_path()` still invalidates the cache when the primary path changes. This is clean, defensive, and production-ready.

### 💡 Suggestions
**File:** `tests/test_camera.py`  
**Lines:** 287, 311

- The docstring update at 287 is clear and well-written. However, the `os.path.exists` patch at 311 is questionable — the test should focus on the fallback behavior without simulating a race condition unless explicitly tested elsewhere. Consider removing the `exists` patch and relying solely on the `open` side-effect simulation.

firstof9:
Please reivew PR comments and shore up test coverage. Thanks!

firstof9 (changes requested):
We need 100% patch coverage.

jasonjhofmann:
Thanks for the reviews @firstof9 @secondof9 — addressed everything in a9fba00 and fb005a6.

**Coverage (90.9% → 100%).** The uncovered line was the `return None` in `async_camera_image` taken when the primary path *is* the bundled placeholder and that file is missing. Added `test_async_camera_image_placeholder_is_primary_missing` to cover that branch. `camera.py` patch coverage is now 100% (287 stmts, 0 missing).

**🔴 Critical — "redundant comparison" guard.** I kept the comparison and removed the *other* clause, because the suggested simplification is inverted. The guard was:

```python
if self._default_image_path and self._file_path != self._default_image_path:
```

- `self._default_image_path` is set unconditionally in `__init__` to a non-empty f-string, so it is *always* truthy — **that** is the genuinely redundant clause, and I removed it.
- `self._file_path != self._default_image_path` is **load-bearing**: for a non-custom camera, `_file_path` and `_default_image_path` are both built from the same `default_image`, so they are equal. When that file is missing, the comparison correctly skips a pointless re-read of the same path and returns `None`. The suggested `if self._default_image_path:` would instead re-read the same missing file and emit a spurious "placeholder also missing" WARNING for the common case.

Simplified to `if self._file_path != self._default_image_path:` with a comment explaining why it is not redundant.

**⚠️ Warning — contradictory `os.path.exists` patch in tests.** Agreed. I verified empirically that **neither** `os.path.exists` nor `os.access` is load-bearing for these read-path tests (the `integration` fixture already mocks the filesystem via `mock_osmakedir`/`mock_listdir`/etc.), so I removed both from all three `async_camera_image` tests. Each now expresses the missing-file condition solely through the `Path(...).open` `FileNotFoundError` side-effect — no more "file exists but is unreadable" contradiction.

All 58 camera tests pass; `ruff check` + `ruff format --check` clean.

secondof9:
## Code Review

### 🔴 Critical
- **custom_components/mail_and_packages/camera.py:148** — Add `self._default_image_path = f"{Path(__file__).parent}/{default_image}"` to the `__init__` method. This captures the bundled placeholder path at camera initialization so it's always available regardless of later `_file_path` changes. This prevents the camera from returning `None` (which causes HTTP 500) when a delivery image is missing on disk.
- **custom_components/mail_and_packages/camera.py:187-215** — Add fallback logic in `_read_file` error handling. When `FileNotFoundError` is raised for the primary image path, the code now falls back to reading the bundled placeholder from `self._default_image_path`. The fallback is skipped when the primary path already equals the placeholder (non-custom cameras), avoiding a pointless second read of the same missing file. On second failure, a `WARNING` is logged and `None` is returned (true edge case).
- **custom_components/mail_and_packages/camera.py:193** — Add debug log message "Could not read camera {self._name} image from file: {self._file_path}; falling back to bundled placeholder {self._default_image_path}" to inform operators that the fallback path is being taken.
- **custom_components/mail_and_packages/camera.py:204-207** — Add guard `if self._file_path != self._default_image_path:` before attempting the fallback read. This prevents double-failure loops on non-custom cameras where both paths point to the same missing file.
- **custom_components/mail_and_packages/camera.py:214-215** — Cache the placeholder bytes against `self._cached_image_path = self._default_image_path` so repeated reads are cheap; `update_file_path()` invalidates the cache when the primary path changes.
- **tests/test_camera.py:312** — Delete old test `test_async_camera_image_file_error` that covered the buggy behavior (returning `None` on missing file). This test must be replaced by tests covering the new fallback contract.
- **tests/test_camera.py:314-321** — Delete two tests that asserted the old buggy behavior (expecting `None` on `FileNotFoundError`). These tests must be replaced.
- **tests/test_camera.py:323-376** — Add new test `test_async_camera_image_falls_back_to_placeholder`. This asserts that when the primary path points at a missing file, the camera returns placeholder bytes instead of `None`, and the cache is keyed to the placeholder path so repeated reads are cheap.
- **tests/test_camera.py:378-404** — Add new test `test_async_camera_image_placeholder_is_primary_missing`. This covers the edge case where the primary path already equals the placeholder (non-custom cameras). In this case, the fallback branch is skipped and `None` is returned, because there's nothing to fall back to.
- **tests/test_camera.py:446-455** — Update existing test `test_camera_file_not_found`. The old test asserted `image is None`, which was the bug. Now it asserts `image is not None` and `len(image) > 0`, verifying the fallback behavior works correctly.
- **tests/test_camera.py:465-476** — Update existing test `test_camera_image_read_error`. The old test simulated a missing file by patching `builtins.open` to raise `FileNotFoundError`. Now it expects the fallback to kick in and return placeholder bytes instead of `None`.

### ⚠️ Warnings
- **custom_components/mail_and_packages/camera.py:148** — The comment above `self._default_image_path` is very long (14 lines). Consider splitting into two comments or moving detailed explanation to a docstring.
- **custom_components/mail_and_packages/camera.py:193-206** — The combined debug message spans 14 lines. Consider breaking into two shorter log messages for better readability.
- **custom_components/mail_and_packages/camera.py:214-215** — The cache assignment comment is very long (7 lines). Consider splitting into two or moving details to a docstring.

### 💡 Suggestions
- **custom_components/mail_and_packages/camera.py:148** — Consider adding a docstring to the `MailCam` class explaining the fallback mechanism and when the cache is invalidated.
- **custom_components/mail_and_packages/camera.py:214-215** — Consider adding a comment explaining why the cache is keyed to the placeholder path (cheap repeat reads vs. re-evaluating the missing-file condition on every request).

### ✅ Looks Good
- **custom_components/mail_and_packages/camera.py** — The fix is minimal, well-scoped, and addresses the root cause (returning `None` causing HTTP 500) without changing the broader architecture.
- **tests/test_camera.py** — New tests cover both the happy path (fallback works) and the edge case (primary already equals placeholder). Existing tests are updated to reflect the new contract.
- **tests/test_camera.py:323-376** — The test `test_async_camera_image_falls_back_to_placeholder` properly mocks the primary path failure and verifies the placeholder bytes are returned.
- **tests/test_camera.py:378-404** — The test `test_async_camera_image_placeholder_is_primary_missing` properly mocks the primary path already pointing at the placeholder and verifies the fallback is skipped.
- **tests/test_camera.py:446-455** — The updated test `test_camera_file_not_found` correctly verifies the fallback behavior.
- **tests/test_camera.py:465-476** — The updated test `test_camera_image_read_error` correctly verifies the fallback behavior.
- **Custom integration** — The integration is stable and well-maintained; this is a targeted fix for a regression (HTTP 500 on missing image).

---

firstof9 (approved):

firstof9:
Thanks!

PR #1279, opened 2026-06-24
fix(amazon): catch the new "has shipped" email subjects

Yesuah:
## Proposed change

Amazon now sends shipped emails with subjects like:

    Your Amazon.com order of "<item>" and 1 more item has shipped!

None of the old AMAZON_SHIPMENT_SUBJECT keywords match that, so the IMAP subject search skipped these emails and they never got counted.

Added "Your Amazon.com order" / "Your Amazon order" to the list, plus a quick test so this doesn't slip through again.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

### DIFF
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -173,6 +173,8 @@ AMAZON_SHIPMENT_SUBJECT = [
     "Shipped:",
     "Enviado:",
     "Out for delivery:",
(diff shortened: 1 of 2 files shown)


Yesuah:
I think it should be 
`has shipped!`

PR #1287, opened 2026-07-01
fix: validate Amazon delivery-image host to prevent SSRF

metaember:
### Summary

`get_amazon_image()` extracts the delivery-photo URL from email HTML and restricts the download to Amazon's S3 bucket. The host is validated using a regex capture group, but that group stops at an `@`. Because email content is attacker-influenced, a crafted `src` such as:

```
https://us-prod-temp.s3.amazonaws.com@example.org/imageId-…?X-Amz-Signature=…
```

passes the check (the captured group equals the bucket) while `aiohttp` actually connects to the host **after** the `@` (the part before `@` is URL *userinfo*, not the host). That lets a delivery email steer the fetch to an arbitrary host — an SSRF from the Home Assistant instance.

### Fix

Parse the assembled URL with `urllib.parse.urlparse` and compare the **real** `hostname`, and reject any URL that carries userinfo. Legit S3 URLs are unaffected; only the smuggled-host case is blocked. One-line-ish change in `helpers.py`, no behavior change for normal emails.

### Tests

- `test_amazon_search_delivered_ssrf` + sample `amazon_delivered_ssrf.eml`: a delivery email whose image URL uses `…s3.amazonaws.com@evil.example/…` must **not** trigger a download.
- The existing `test_amazon_search_delivered` (legit S3 image *does* download) still passes.

Verified locally that the sample email matches the existing `AMAZON_IMG_PATTERN` (so it exercises the vulnerable path) and that the new host check rejects it while accepting the legitimate S3 URL.

### DIFF
--- a/custom_components/mail_and_packages/helpers.py
+++ b/custom_components/mail_and_packages/helpers.py
@@ -15,6 +15,7 @@ from datetime import timezone
 from email.header import decode_header
 from shutil import copyfile, copytree, which
 from typing import Any, List, Optional, Type, Union
+from urllib.parse import urlparse
 
 import aiohttp
 import imageio as io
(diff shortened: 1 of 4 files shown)


firstof9:
Please review [CONTRIBUTING.md](https://github.com/moralmunky/Home-Assistant-Mail-And-Packages/blob/master/CONTRIBUTING.md)

PR #1297, opened 2026-07-09
fix(amazon): filter out delivered orders from shipped orders list

### LINKED ISSUE #1296: ISSUE: order attribute on sensor.mail_amazon_packages includes already-delivered orders — should only show in-transit orders
**Describe the bug**

The `order` attribute on `sensor.mail_amazon_packages` returns all Amazon order numbers found in shipped emails within the lookback window, including orders that have already been delivered. It should only return order numbers for packages that are currently in transit, consistent with how the package count itself works and how FedEx/UPS/USPS tracking numbers behave on their respective delivering sensors.

**Root cause (code analysis)**

In `shippers/amazon.py`, `_parse_amazon_emails` maintains two separate context collections:

```python
context = {
    "all_shipped_orders": set(),   # populated by _handle_shipping_email
    "delivered_packages": {},       # populated by _handle_delivered_email
    ...
}
```

`_calculate_final_count` correctly subtracts delivered packages from arriving packages to produce an accurate in-transit count:

```python
for order_id, arriving_count in ctx["packages_arriving_today"].items():
    delivered_count = ctx["delivered_packages"].get(order_id, 0)
    final_count += max(0, arriving_count - delivered_count)
```

However when `param == "order"`, the function returns `list(context["all_shipped_orders"])` without applying the same delivered subtraction:

```python
if param == "count":
    return final_count
return list(context["all_shipped_orders"])  # <-- never filtered against delivered_packages
```

This means with a 3-day lookback window, any order shipped in the last 3 days appears in the `order` attribute even if it was delivered yesterday. The count and the order list are inconsistent — the count correctly shows 0 in-transit packages while the `order` attribute still lists the delivered order numbers.

**The fix**

Filter `all_shipped_orders` against `delivered_packages` before returning, matching the same logic used in `_calculate_final_count`:

```python
if param == "count":
    return final_count

# Only return orders that haven't been fully delivered
in_transit_orders = [
    order_id for order_id in context["all_shipped_orders"]
    if context["packages_arriving_today"].get(order_id, 0)
    > context["delivered_packages"].get(order_id, 0)
]
return in_transit_orders
```

**Environment:**
- OS: HassOS
- Type of device: Raspberry Pi 5
- Home Assistant version: 2026.6.4
- Hassio/Docker/Core: Hassio
- Component version: 0.5.10
- Email provider: Gmail via IMAP (OAuth)

**Steps to reproduce**
1. Install 0.5.10 with Amazon Packages enabled and lookback window set to 3 days
2. Order an Amazon item, receive it same-day
3. The following day, check attributes of `sensor.mail_amazon_packages` — the delivered order number still appears in the `order` attribute despite the package being delivered
4. Confirm `sensor.mail_amazon_packages` state is 0 (count is correct) while `order` still lists the delivered order number (inconsistent)

**Expected behavior**

The `order` attribute should only contain order numbers for packages that are currently in transit — matching the behavior of the package count and consistent with how `tracking_#` works on FedEx/UPS/USPS delivering sensors.

**Observed behavior**

The `order` attribute lists all order numbers from shipped emails within the lookback window regardless of delivery status. With a 3-day window, delivered orders continue appearing in the attribute for up to 3 days after delivery.

**Logs**

Not required — the behavior is directly traceable to `list(context["all_shipped_orders"])` being returned without filtering against `context["delivered_packages"]`.

**Additional context**

The data needed to correctly filter the order list is already present in the context dict — `delivered_packages` is populated correctly by `_handle_delivered_email`. This is purely an omission in the return logic, not a data collection problem.

firstof9:
## Proposed change

Currently, the `order` attribute on `sensor.mail_amazon_packages` lists all Amazon order numbers found in shipped emails within the lookback window, including orders that have already been delivered.

This PR filters the returned order list against `context["delivered_packages"]` so that only in-transit orders (either arriving today and not yet delivered, or shipped and not yet delivered) are displayed, matching the package count logic and resolving the inconsistency.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [x] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1296

### DIFF
--- a/custom_components/mail_and_packages/shippers/amazon.py
+++ b/custom_components/mail_and_packages/shippers/amazon.py
@@ -210,7 +210,17 @@ class AmazonShipper(Shipper):
 
         if param == "count":
             return final_count
-        return list(context["all_shipped_orders"])
+
+        return [
+            order_id
+            for order_id in context["all_shipped_orders"]
+            if context["packages_arriving_today"].get(order_id, 0)
+            > context["delivered_packages"].get(order_id, 0)
+            or (
+                context["packages_arriving_today"].get(order_id, 0) == 0
+                and context["delivered_packages"].get(order_id, 0) == 0
+            )
+        ]
 
     async def _process_amazon_email(
         self,
--- a/tests/shippers/test_amazon.py
+++ b/tests/shippers/test_amazon.py
@@ -648,6 +648,42 @@ async def test_amazon_order_list(hass, mock_imap_amazon_shipped):
 
 
 @pytest.mark.asyncio
+async def test_amazon_order_list_filtering(hass, mock_imap_amazon_shipped):
+    """Test Amazon order list filtering when delivered."""
+    shipper = AmazonShipper(hass, {})
+    with (
+        patch(
+            "custom_components.mail_and_packages.shippers.amazon.email_fetch",
+            new_callable=AsyncMock,
+            side_effect=[
+                (
+                    "OK",
+                    [
+                        b"Header",
+                        b"Subject: Shipped:\n\nYour order 111-1234567-1234567 has shipped.",
+                    ],
+                ),
+                (
+                    "OK",
+                    [
+                        b"Header",
+                        b"Subject: Delivered: Your Amazon order has arrived!\n\nYour order 111-1234567-1234567 was delivered.",
+                    ],
+                ),
+            ],
+        ),
+        patch(
+            "custom_components.mail_and_packages.shippers.amazon.search_amazon_emails",
+            new_callable=AsyncMock,
+            return_value=[b"1", b"2"],
+        ),
+    ):
+        result = await shipper.process(mock_imap_amazon_shipped, "today", AMAZON_ORDER)
+        # The order was shipped and delivered, so it should be filtered out
+        assert "111-1234567-1234567" not in result[AMAZON_ORDER]
+
+
+@pytest.mark.asyncio
 async def test_amazon_hub_multi(hass, mock_imap_amazon_the_hub):
     """Test Amazon Hub with multiple codes and deduplication."""
     shipper = AmazonShipper(hass, {})


secondof9:
## Code Review — PR #1297: fix(amazon): filter out delivered orders from shipped orders list

### 🔍 Architectural Walkthrough

**`custom_components/mail_and_packages/shippers/amazon.py`**

The PR modifies `_parse_amazon_emails()`, specifically the `param == "order"` return path. Previously it returned `list(context["all_shipped_orders"])` — every order found in shipped emails, including ones already delivered. The new code filters this list to only include orders where `packages_arriving_today > delivered_packages` (arriving today, not yet delivered) or `packages_arriving_today == 0 and delivered_packages == 0` (not arriving, not yet delivered — i.e., new orders).

The filter uses defensive `.get(order_id, 0)` on both context dicts, which is appropriate since these dicts are populated by `process()` during its email loop, but the defensive pattern prevents breakage if any edge case produces a missing key.

**`tests/shippers/test_amazon.py`**

New test `test_amazon_order_list_filtering` mocks two emails: one shipped, one delivered for the same order. It asserts the delivered order is filtered out. The test correctly exercises the intended behavior.

### ✅ Looks Good

- **Correct fix for reported bug** — The PR description says orders appearing in both shipped and delivered lists is the issue. The filter logic directly addresses this by requiring `packages_arriving_today > delivered_packages` for inclusion.
- **Defensive dict access** — `.get(order_id, 0)` on both `context["packages_arriving_today"]` and `context["delivered_packages"]` prevents `KeyError` if either dict is missing a key.
- **Clean test coverage** — The test correctly sets up a shipped-then-delivered scenario and asserts the expected outcome.
- **No behavioral impact on other parameters** — The filter only applies to the `order` return path. The `count`, `packages_arriving_today`, and `packages_arriving_today_keys` paths are unaffected.
- **HA architectural compliance** — No blocking I/O, no coordination changes, pure data transformation on existing context.

### ⚠️ Warning

- **`all_shipped_orders` semantics** — The context key `all_shipped_orders` contains order IDs that appear in any shipped email within the lookback window. The filter assumes this includes delivered orders (which it does, since delivered emails also contain the order number). The filter correctly excludes these, but it's worth noting that `all_shipped_orders` is a misnomer — it's really "all order IDs found in any email in the window." This is a cosmetic concern; the code is functionally correct.

### 💡 Suggestions

- The filter could be slightly simplified using a single comparison: `context["packages_arriving_today"].get(order_id, 0) > context["delivered_packages"].get(order_id, 0)`. This would exclude orders where `arriving_today <= delivered` — which covers both fully-delivered orders AND orders that haven't shipped yet (arriving=0, delivered=0). The current two-condition approach adds clarity but has the same effect. Minor.

---

**Assessment: Approved — the change correctly fixes the reported inconsistency where delivered Amazon orders appeared in the shipped order list.**

PR #1299, opened 2026-07-10
fix(coordinator): zero delivering count when all tracked packages are delivered

jasonjhofmann:
## Proposed change

`<carrier>_delivering` sensors get stuck at the raw IMAP email count whenever **every** package in the days-back window has already been delivered — the exact case the delivered-dedup machinery exists for. Two gaps compound:

1. **Batch dedup used the today-only delivered list.** `_deduplicate_batch_tracking` builds its delivered set from the `_delivered` sensor's `ATTR_TRACKING`, which `process()` deliberately overwrites with today-only results so the sensor resets at midnight. Packages delivered on any *previous* day inside the window were therefore never subtracted from `_delivering`. It now uses `pre_filtered_tracking` (the extended-window list captured before the today-only overwrite) when present.

2. **The coordinator's count override skipped the empty-map case.** `_apply_tracking_state` only overrode `<prefix>_delivering` when the in-transit map was non-empty. When delivered notifications removed every tracking number, the override was skipped, the un-deduped batch count leaked through as the sensor value, and the `<prefix>_tracking` attribute vanished. The override now also fires for carriers that reported **delivering/exception** tracking details this scan (defense in depth at the state-machine layer). Delivered-only details deliberately do **not** trigger it: a carrier whose delivering emails yield no extractable tracking numbers has a legitimate email-based count that tracking-level dedup cannot verify, and zeroing it would be a regression.

Real-world repro (how this was found): 12 FedEx packages delivered over a week stayed "out for delivery" on the dashboard; once their delivered emails were all present in the scanned folder, the count *snapped back to the raw 12* because the emptied in-transit map skipped the override. With this change the first scan reports 0.

The all-delivered transition is logged at debug so the raw→0 override is visible in diagnostics.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Tests: 4 new regression tests — all-delivered zeroes the count (fails on `dev`), extended-window dedup subtracts prior-day deliveries at batch level (fails on `dev`), delivered-only tracking details preserve an email-based count, and carriers with no tracking details keep their email counts. Full suite passes (564), ruff/mypy/pre-commit clean.

### DIFF
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -365,9 +365,34 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
             )
 
             in_transit = self._in_transit_tracking.get(prefix, {})
-            if in_transit:
+            # A carrier that reported DELIVERING/EXCEPTION tracking details
+            # this scan must have its count overridden even when the
+            # in-transit map ends up EMPTY: when every tracked package has a
+            # delivered notification, the raw IMAP count (which cannot dedup
+            # prior-day deliveries) would otherwise leak through as the
+            # sensor value. Batch-level dedup already zeroes the count for
+            # shippers that emit tracking details, so this is defense in
+            # depth at the state-machine layer. Delivered-only details must
+            # NOT trigger the override: a carrier whose delivering emails
+            # yielded no extractable tracking numbers has a legitimate
+            # email-based count that tracking-level dedup cannot verify —
+            # and carriers with no tracking details at all keep their
+            # email-count value untouched.
+            has_details = any(
+                f"{prefix}_{suffix}" in tracking_details
+                for suffix in ("delivering", "exception")
+            )
+            if in_transit or has_details:
+                if not in_transit and data.get(f"{prefix}_delivering"):
+                    _LOGGER.debug(
+                        "Prefix '%s': no tracked packages remain in transit — "
+                        "overriding delivering count %s -> 0",
+                        prefix,
+                        data.get(f"{prefix}_delivering"),
+                    )
                 data[f"{prefix}_tracking"] = list(in_transit.keys())
                 data[f"{prefix}_delivering"] = len(in_transit)
+            if in_transit:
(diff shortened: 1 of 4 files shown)


secondof9 (commented):
## Hermes Agent Code Review

### Summary

Clean bugfix with excellent defensive programming. Two complementary changes fix the exact regression where all-delivered packages would stick at the raw IMAP count: (1) the batch dedup now uses the extended-window tracking list, and (2) the coordinator override fires even when the in-transit map is empty.

### Strengths

- **Clear problem statement** in the docstrings: `_delivered` tracking resets at midnight, so prior-day deliveries need `pre_filtered_tracking`
- **Defense in depth**: the batch-level fix handles the primary case, and the coordinator override adds a second layer
- **Precise filtering**: `has_details` only matches `delivering`/`exception` suffixes, deliberately excluding `delivered`-only entries
- **4 well-structured regression tests** covering: extended-window dedup, all-delivered zeroing, delivered-only preservation, and no-details preservation
- **Debug logging** on the override path for diagnostics

### Minor Notes

1. **`coordinator.py:388-390`** — The `_packages` field is intentionally NOT recomputed when `in_transit` is empty after the override. The test docstring explains this well, but this means `_packages` will never be zeroed for an all-delivered carrier. The batch dedup already made it the shipped-but-not-yet-out count, and the comment confirms this is correct. If someone reads the code without the docstring, the absence of recalculation might look like a bug. A one-line comment on the `if in_transit:` gate would make this clearer.

2. **`tests/test_coordinator.py:389`** — Same design point: `data["fedex_packages"] == 2` is asserted even when all 2 packages are delivered. This is the same deliberate choice as #1.

### Verdict

This is a well-targeted fix for a real regression. The two changes complement each other perfectly, the tests cover the new behavior and the edge cases, and the code is clear. The only two points above are readability suggestions, not blockers.

---
*Reviewed by Hermes Agent (secondof9)*

firstof9:
Thanks!

PR #1300, opened 2026-07-10
fix(camera): never write or serve 0-byte delivery images

jasonjhofmann:
## Proposed change

FedEx delivered notifications (and potentially other carriers' lazy-loaded HTML) can contain a **bare/empty base64 data URI** (`data:image/jpeg;base64,` with no payload). The extraction regex's capture group happily matches zero characters, `b64decode("")` returns `b""`, and a 0-byte "delivery photo" is written to disk. HA core's camera proxy treats empty bytes like a missing image and serves **HTTP 500**, so the camera entity renders as a broken image on dashboards.

Three layers, outermost first:

- **`_extract_from_html`** skips empty base64 matches, so a spacer data URI earlier in the HTML can no longer shadow a real photo later in the same part (previously `matches[0]` was taken blindly and the whole HTML pass returned `False`).
- **`save_image_data_to_disk`** rejects empty/`None` payloads and reports failure, so extraction falls through to the next pass instead of "succeeding" with nothing. This also fixes a latent `TypeError` when a CID part's `get_payload(decode=True)` returns `None` (`write(None)` was not caught by the `except OSError`).
- **The camera's `_read_file`** treats a 0-byte file like a missing one, routing through the existing bundled-placeholder fallback (#1268) — covering files already on disk from before this fix.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR is related to issue: follows up #1268 (placeholder fallback for missing images)

Tests: empty-payload rejection (fails on `dev`), 0-byte-file placeholder fallback with read-order assertions (fails on `dev`), and empty-match-before-real-photo recovery. The pre-existing decode-error test was updated to use a well-formed base64 payload so it still exercises the `b64decode` error branch now that empty matches are filtered. Full suite passes (563), ruff/mypy/pre-commit clean.

### DIFF
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -182,7 +182,13 @@ class MailCam(CoordinatorEntity, Camera):
 
         def _read_file(path: str) -> bytes:
             with Path(path).open("rb") as f:
-                return f.read()
+                data = f.read()
+            if not data:
+                # A 0-byte image (e.g. a failed extraction that wrote an empty
+                # file) is as unservable as a missing one — raise so it routes
+                # through the same placeholder fallback below.
+                raise FileNotFoundError(f"empty image file: {path}")
+            return data
 
         try:
             image_bytes = await self.hass.async_add_executor_job(
--- a/custom_components/mail_and_packages/utils/shipper.py
+++ b/custom_components/mail_and_packages/utils/shipper.py
@@ -114,8 +114,21 @@ def _find_tracking_in_body(
     return None
 
 
-def save_image_data_to_disk(shipper_name: str, path: str, image_data: bytes) -> bool:
+def save_image_data_to_disk(
+    shipper_name: str, path: str, image_data: bytes | None
+) -> bool:
     """Write image bytes to disk and verify."""
+    if not image_data:
+        # Extraction can hand us zero bytes (e.g. a bare/empty base64 data URI
+        # in the email HTML). Writing that produces a 0-byte "photo" the
+        # camera then serves as a broken image — report failure instead so
+        # the caller falls through to the next extraction pass.
+        _LOGGER.debug(
+            "%s - No image data extracted; not writing %s",
+            shipper_name,
+            path,
+        )
+        return False
     try:
         # Ensure directory exists
         directory = Path(path).parent
@@ -240,8 +253,11 @@ def _extract_from_html(
             else str(payload)
         )
 
-        # Base64 check
-        if matches := re.findall(base64_pattern, content):
+        # Base64 check. Skip empty matches: a bare "data:image/...;base64,"
+        # URI (seen in real FedEx delivered emails) matches zero characters,
+        # and taking it would discard a real photo later in the same part.
+        matches = [m for m in re.findall(base64_pattern, content) if m]
+        if matches:
             try:
                 base64_data = matches[0].replace(" ", "").replace("=3D", "=")
                 return save_image_data_to_disk(
--- a/tests/test_camera.py
+++ b/tests/test_camera.py
@@ -1,7 +1,7 @@
 """Tests for camera component."""
 
 from pathlib import Path
-from unittest.mock import AsyncMock, MagicMock, mock_open, patch
+from unittest.mock import AsyncMock, MagicMock, call, mock_open, patch
 
 import pytest
 from homeassistant.const import ATTR_ENTITY_ID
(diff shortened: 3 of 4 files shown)


secondof9:
## Code Review — PR #1300: fix(camera): never write or serve 0-byte delivery images

### 🔴 Critical Issues: None

### ⚠️ Warnings: None

### 💡 Suggestions: None

### ✅ Approved

---

## Line-by-Line Analysis

### `custom_components/mail_and_packages/camera.py` (+7 / -1)

**Lines 182-190** — `_read_file` helper inside `async_camera_image`:
```python
def _read_file(path: str) -> bytes:
    with Path(path).open("rb") as f:
        data = f.read()
    if not data:
        # A 0-byte image ... routes through the same placeholder fallback below.
        raise FileNotFoundError(f"empty image file: {path}")
    return data
```

- **Good**: Reads into a variable first, then validates. The `raise FileNotFoundError` is the correct signal — the caller catches this and falls through to the bundled placeholder fallback (PR #1268 path). The type hint `-> bytes` remains accurate because when this function reaches `return data`, `data` is guaranteed non-empty bytes.
- **Minor note**: The `if not data:` check catches `b""` (empty bytes) but `f.read()` never returns `None` for a valid opened file object. This is correct behavior.
- **Async safety**: `_read_file` is called inside `await self.hass.async_add_executor_job(...)`. No blocking I/O leaks.

### `custom_components/mail_and_packages/utils/shipper.py` (+19 / -3)

**Change 1 — `save_image_data_to_disk` signature (~line 114)**:
```python
def save_image_data_to_disk(
    shipper_name: str, path: str, image_data: bytes | None
) -> bool:
```

- **Good**: Signature expanded to accept `bytes | None`. The `if not image_data:` guard catches both `None` and `b""`. The `_LOGGER.debug(...)` call is well-placed and informative. Returns `False` so the caller falls through to the next extraction pass.
- **Defensive**: This prevents a `TypeError` if a CID part's `get_payload(decode=True)` returns `None` (the old `except OSError` wouldn't have caught that).

**Change 2 — `_extract_from_html` base64 filtering (~line 253)**:
```python
matches = [m for m in re.findall(base64_pattern, content) if m]
if matches:
```

- **Excellent design**: Filters empty regex matches *before* taking `matches[0]`. This prevents a bare `data:image/jpeg;base64,` URI (with zero-length capture group) from shadowing a real photo later in the HTML. Previously `matches[0]` would be `""`, `b64decode("")` returns `b""`, and a 0-byte file gets written — triggering the HTTP 500 bug.
- **Correctness**: The list comprehension is idiomatic Python and avoids an unnecessary `if m:` per match overhead (which is fine for this use case given the small size of email HTML parts).

**Change 3 — Test data update**:
```python
sdata = b'Content-Type: text/html\n\n<html><body><img src="data:image/jpeg;base64,YWJjZA=="></body></html>'
```

- **Good**: Replaced `bad-base64!` with well-formed base64 `YWJjZA==` ("abcd"). The old test would produce an empty match with the new filtering logic, bypassing the `b64decode` error-handling branch entirely. The new test correctly reaches the mocked `TypeError` branch.

### `tests/test_camera.py` (+58 / -1)

**New test — `test_async_camera_image_empty_file_falls_back_to_placeholder`**:
- Well-structured regression test for the camera's 0-byte file handling.
- Uses `MagicMock` hierarchy for both primary and fallback `Path.open()` calls.
- **Key assertions**: Tests not just that the fallback works, but also that the call order is correct — the primary (empty) file is opened *first*, then the placeholder. This catches any implementation that incorrectly re-reads the primary for the fallback path.
- **Fixture cleanliness**: Doesn't interfere with other tests; uses its own `mock_path_class` patch.

### `tests/utils/test_shipper.py` (+50 / -2)

**New test — `test_save_image_data_to_disk_rejects_empty_data`**:
- Verifies both `b""` and `None` inputs are rejected.
- Asserts `Path().open` is never called for empty data.
- Verifies the debug log message is emitted.

**New test — `test_extract_from_html_skips_empty_base64_before_real_image`**:
- Tests the core fix: a bare `data:image/jpeg;base64,` spacer before a real photo must not shadow the real image.
- Mocks `save_image_data_to_disk` with `return_value=True` to verify the extraction proceeds past the empty match.
- Asserts the *real* photo bytes are extracted, not the empty match.

**Updated test — `test_extract_from_html_decode_error`**:
- Correctly updated to use well-formed base64 so the test still exercises the `b64decode` error path.

---

## Architecture Assessment

- **HA Core compliance**: Camera proxy fallback is integrated at the right layer. No violation of the raw API / frontend separation principle.
- **Async safety**: All file I/O goes through `hass.async_add_executor_job`. No blocking calls in the async path.
- **Error boundaries**: `FileNotFoundError` from `_read_file` is caught and cleanly routed to the placeholder. No unhandled exceptions that could crack the loop.
- **Test coverage**: 2 new tests in `test_shipper.py`, 1 new test in `test_camera.py`, 1 test updated. Total of 4 new/updated test functions covering all three defense layers. All passing (563 total per the PR description).
- **Code quality**: Clean, idiomatic Python. No over-engineering. Docstrings explain the *why* at each layer, not just the *what*.

---

## Summary

This is a clean, well-designed bugfix PR that fixes a real bug (0-byte delivery images causing HTTP 500 in the HA camera proxy) with three complementary defense layers:

1. **Extraction layer** (`shipper.py`): Filters empty regex matches so bare data URIs can't shadow real photos.
2. **Write layer** (`shipper.py`): Rejects empty/None payloads so 0-byte files are never written.
3. **Serve layer** (`camera.py`): Treats 0-byte files identically to missing files, routing to the placeholder fallback.

All layers are properly tested with regression tests. No critical issues found.

firstof9:
Thanks!

PR #1301, opened 2026-07-10
fix(imap): percent-encode folder names in multi-folder composite IDs

jasonjhofmann:
## Proposed change

Multi-folder searches tag each UID with its source folder as a composite `folder/uid` ID. Those composite IDs are space-joined (mimicking a raw IMAP SEARCH response) and re-split on whitespace by every consumer — so a folder name containing spaces, which is entirely legal in IMAP, **shatters into garbage IDs**: `My Folder/55` becomes `My`, `Folder/55`. Subject verification then drops the fragments, tracking extraction misses those emails, and delivered-dedup silently degrades for anything filed outside single-token folders.

This percent-encodes the folder component (`urllib.parse.quote(..., safe="")`) at both producers (`_parse_esearch_line` for the ESEARCH path, and the sequential per-folder fallback in `_execute_single_search`) and decodes it at all four consumers (`email_fetch`, `email_fetch_headers`, `email_fetch_text`, `email_fetch_batch`). The encoded form contains no whitespace and no `/`, so both the `.split()` and the `rsplit("/", 1)` round-trips are lossless for any folder name — including names containing literal `%` sequences, `/`, tabs, or non-ASCII (UTF-7 wire names are decoded to real Unicode *before* encoding, and `selectfolder` re-encodes at select time, unchanged).

Kept as an internal ID-format change rather than restructuring the space-joined format itself: that format is consumed at ~14 call sites across the shippers and would balloon the diff.

Note: debug logs of matched email IDs now show the percent-encoded folder form.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Tests: encode/decode round-trip property test (spaces, `%`, `/`, tabs, non-ASCII), spaced-folder end-to-end tests for both search paths (ESEARCH and sequential fallback) proving the joined blob re-splits into exactly one ID per email (fail on `dev`), and decoded-folder-selection tests for all four fetch consumers. Full suite passes (567), ruff/mypy/pre-commit clean.

### DIFF
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -5,6 +5,7 @@ import binascii
 import logging
 import re
 import unicodedata
+from urllib.parse import quote, unquote
 
 import aioimaplib
 from aioimaplib import (
@@ -113,6 +114,30 @@ def quote_folder(folder: str) -> str:
     if folder.startswith('"') and folder.endswith('"'):
         return folder
     return folder if _is_imap_atom(folder) else f'"{folder}"'
+
+
+def encode_folder_ref(folder: str) -> str:
+    """Percent-encode a folder name for use in a composite ``folder/uid`` ID.
+
+    Multi-folder searches tag each UID with its source folder as
+    ``folder/uid``. Those composite IDs are space-joined and re-split on
+    whitespace at several call sites, and split on ``/`` to recover the
+    folder — so the folder component must contain neither whitespace nor
+    ``/``. A folder named ``# - Projects`` would otherwise shatter into
+    ``#``, ``-``, ``Projects/55`` when the joined ID list is ``.split()``.
+    ``quote(..., safe="")`` escapes both (and ``%`` itself, keeping the
+    round-trip lossless for any folder name).
+    """
+    return quote(folder, safe="")
+
+
+def decode_folder_ref(folder: str) -> str:
+    """Decode the percent-encoded folder component of a composite ID.
+
+    Takes the already-split folder component (everything before the final
+    ``/`` of a ``folder/uid`` ID), not the full composite ID.
+    """
+    return unquote(folder)
 
 
 class InvalidAuth(HomeAssistantError):
@@ -382,7 +407,7 @@ def _parse_esearch_line(line_bytes: bytes) -> list[bytes]:
                 pass
         else:
             uids.append(part)
-    return [f"{mailbox}/{uid}".encode() for uid in uids]
+    return [f"{encode_folder_ref(mailbox)}/{uid}".encode() for uid in uids]
 
 
 async def _execute_single_search(account: IMAP4_SSL, search_query: str) -> list[bytes]:  # noqa: C901
@@ -446,7 +471,8 @@ async def _execute_single_search(account: IMAP4_SSL, search_query: str) -> list[
                 if res.result == "OK" and res.lines:
                     parsed = parse_search_response(res.lines)
                     all_uids.extend(
-                        f"{folder}/{uid.decode()}".encode() for uid in parsed
+                        f"{encode_folder_ref(folder)}/{uid.decode()}".encode()
+                        for uid in parsed
                     )
             except TimeoutError:
                 raise
@@ -570,7 +596,7 @@ async def email_fetch(account: IMAP4_SSL, num, parts: str = "(RFC822)") -> tuple
     num_str = num.decode() if isinstance(num, bytes) else str(num)
     if "/" in num_str:
         folder, num_str = num_str.rsplit("/", 1)
-        await selectfolder(account, folder)
+        await selectfolder(account, decode_folder_ref(folder))
         try:
             res = await account.uid("FETCH", num_str, parts)
         except TimeoutError:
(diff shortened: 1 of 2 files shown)


secondof9 (approved):
## Line-by-Line Code Review: PR #1301

### 🛠️ Summary

This PR fixes a real, under-the-radar IMAP composite ID splitting bug where folder names containing spaces shatter into garbage IDs during multi-folder search. The fix is a clean 3-step encode-at-produce, split-and-decode-at-consume pattern applied consistently across both search paths and all four email fetch consumers. The design is correct: percent-encoding with `safe=""` produces a minimal, whitespace-free, slash-free token form that round-trips losslessly through `.split()` and `rsplit("/", 1)` at every call site.

### ✅ What's Excellent

1. **`encode_folder_ref` / `decode_folder_ref` pairing** — The new functions at lines 116-140 are pure, zero-state helpers. No IMAP state, no session coupling. They compose perfectly.

2. **Both producer sites are covered** — `_parse_esearch_line` (line 382 → 407) and the sequential fallback `_execute_single_search` (line 446 → 471) both encode at write. No code path left behind.

3. **All four consumers decode** — `email_fetch`, `email_fetch_headers`, `email_fetch_text`, and `email_fetch_batch` each call `decode_folder_ref(folder)` before `selectfolder`. The batch function's grouping logic (`folder_to_nums.setdefault(folder, [])`) now receives decoded folder keys, so emails group correctly.

4. **Test coverage is surgical** — `test_folder_ref_roundtrip` uses 8 diverse folder names covering spaces, dashes, percent signs, slashes, tabs, and UTF-8 non-ASCII. The 6 async regression tests each target one specific consumer, with precise mock assertions.

5. **No architecture drift** — Kept as an internal ID-format change rather than restructuring the space-joined composite format. Touching ~14 call sites would have been over-engineering.

6. **`%` self-encoding** — `quote("", safe="")` escapes `%` itself, which is critical. A folder named `50% discount codes` becomes `50%25%20discount%20codes`, which `unquote` decodes back exactly. This property is correctly exercised in the round-trip test.

### ⚠️ Warnings (Non-blocking)

**1. `quote_folder` function is now orphaned (line 115).**
The existing `quote_folder` function (lines 115-117) is the IMAP-level quoting for folder names sent to `SELECT`/`LIST` commands. It handles the IMAP atom check (`_is_imap_atom`) and wraps non-atoms in quotes. This is *different* from `encode_folder_ref`, which is about encoding for composite IDs. The two don't interfere, but it's worth noting that both exist in the same file and solve different problems. The diff correctly doesn't touch `quote_folder` — good call.

**2. Test mocks don't verify `has_capability` return on `test_email_search_multisearch_spaced_folder`.**
Line 1176: `mock_account.has_capability = MagicMock(return_value=True)`. This is actually fine — for the ESEARCH test we want `has_capability` to return `True` so it uses the ESEARCH path. But a reviewer might not immediately see why it's `True` here vs `False` in the sequential test. The assertion that ESEARCH is triggered would be clearer with a comment or by asserting `mock_account.protocol.execute.assert_called_once()` in the ESEARCH test. Not a bug, but a readability concern.

**3. `email_fetch_batch` mock has `host = "imap.gmail.com"` which is dead state.**
Lines 1207-1211: Setting `mock_account.host = "imap.gmail.com"` serves no purpose — it's never read by `email_fetch_batch`. Minor dead-state leak in a test.

### 💡 Suggestions

**1. Consider `safe="/+"` for `encode_folder_ref` to avoid encoding forward slashes in UTF-7 wire names.**
The PR description notes "UTF-7 wire names are decoded to real Unicode before encoding." This means folder names as *stored* on the server already have `/` resolved. The encoding layer only needs to handle what `unquote` produced. `safe=""` is aggressive but correct for this case. However, if a future IMAP client stores multi-byte UTF-7 wire names with unencoded `/` characters (unlikely but possible with edge-case servers), `safe="/+"` would preserve forward slashes while still encoding spaces. A conservative reviewer flag: `safe=""` is fine *today*, but `safe="/+"` would be more future-proof without breaking current behavior.

**2. The `email_fetch_batch` test could assert `uid.return_value` shape to catch fetch failures.**
Line 1220: The test mocks `mock_res = MagicMock(result="OK", lines=[b"RFC822", b"body"])`. If a real email failed to fetch, we'd want to know. The test currently doesn't assert that `uid` was called with the right number of UIDs (55 and 56). Adding `mock_account.uid.assert_called_once_with("FETCH", "55,56", "(RFC822)")` is already there — actually, wait, line 1221 does assert exactly that. So the test is complete.

### ✅ Looks Good

- **Line 65-77**: `encode_folder_ref` is a pure function, <20 lines, no side effects, clear docstring with a working example.
- **Line 80-85**: `decode_folder_ref` — equally pure, zero boilerplate.
- **Line 382**: Producer site for ESEARCH — `encode_folder_ref(mailbox)}/{uid}` correctly encodes the mailbox name.
- **Line 446**: Producer site for sequential fallback — `encode_folder_ref(folder)}/{uid.decode()}` correctly encodes the folder. Note the indented block is clean: `all_uids.extend(f"{encode_folder_ref(folder)}/{uid.decode()}".encode() for uid in parsed for uid in parsed` — actually the indentation in the diff shows the `for uid` loop is moved into the list comprehension scope, which is correct Python.
- **Lines 570, 597, 627**: All three `email_fetch*` functions decode before calling `selectfolder`. Each independently handles the composite ID split and decodes the folder component.
- **Line 688**: `email_fetch_batch` decodes the folder, then uses the decoded value for grouping in `folder_to_nums.setdefault(folder, [])`. This ensures emails from the same folder (regardless of encoding state in the composite ID) are batched together.
- **Lines 1127-1288**: All new tests are well-structured with clear docstrings explaining the regression they guard against.

### Final Verdict

This is a well-executed, surgical fix. The encode/decode pair is minimal, the producers and consumers are fully covered, and the test suite is comprehensive. No blocking issues. Minor cosmetic suggestions only.

firstof9:
Thanks!

PR #1302, opened 2026-07-10
fix(generic): run delivery-image extraction off the event loop

jasonjhofmann:
## Proposed change

`generic_delivery_image_extraction` parses the full email body and writes the extracted delivery photo with blocking file I/O (`Path.open("wb")`), but is called directly from the async coordinator path. Home Assistant's blocking-call detection flags it on every image extraction:

```
Detected blocking call to open with args (PosixPath('custom_components/mail_and_packages/images/fedex/….jpg'), 'wb')
inside the event loop by custom integration 'mail_and_packages' at
custom_components/mail_and_packages/utils/shipper.py: with Path(path).open("wb") as the_file
```

This dispatches the whole sync extraction through `hass.async_add_executor_job`, matching how the surrounding code already offloads directory creation, placeholder copies, `copy_images`, and the camera's file reads. Exception semantics are unchanged (`OSError` is handled inside the sync function; anything else propagates through the awaited executor future to the same handler in `_update_shippers` as before).

One theoretical note: with the write off the loop, two config entries sharing the default image path *and* default image name could overlap writes to the same file from two executor threads (last-writer-wins becomes a torn-write window). This needs 2+ IMAP accounts with the same carrier delivering during overlapping refreshes; flagging for completeness — a temp-file + `os.replace` write would close it if desired.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Tests: a thread-identity regression test records which thread each extraction call runs on and asserts none is the event-loop thread (fails on current `dev`, passes with the fix). The probe is installed with `patch(..., new=<plain function>)` deliberately — the test harness's `async_add_executor_job` wrapper runs Mock targets inline on the loop, which would defeat the check. Full suite passes (561), ruff/mypy/pre-commit clean.

### DIFF
--- a/custom_components/mail_and_packages/shippers/generic.py
+++ b/custom_components/mail_and_packages/shippers/generic.py
@@ -662,7 +662,11 @@ class GenericShipper(Shipper):
                 msg_parts = (await email_fetch(account, eid, "(RFC822)"))[1]
             for response_part in msg_parts:
                 if isinstance(response_part, (bytes, bytearray)):
-                    if generic_delivery_image_extraction(
+                    # The extraction does blocking file I/O (the image
+                    # write) and CPU-heavy email parsing — run the whole
+                    # sync function off the event loop.
+                    if await self.hass.async_add_executor_job(
+                        generic_delivery_image_extraction,
                         response_part,
                         s_config["image_path"],
                         s_config["image_name"],
--- a/tests/shippers/test_generic.py
+++ b/tests/shippers/test_generic.py
@@ -1,5 +1,6 @@
 """Tests for generic shipper utilities."""
 
+import threading
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


secondof9:
## Code Review — PR #1302

### Summary
Fix: Move blocking `generic_delivery_image_extraction` call off the event loop by wrapping with `hass.async_add_executor_job`.

### Line-by-line review

#### `generic.py` (lines 662–672)

**Original:**
```python
if generic_delivery_image_extraction(
    response_part,
    s_config["image_path"],
    s_config["image_name"],
```

**Fixed:**
```python
# The extraction does blocking file I/O (the image
# write) and CPU-heavy email parsing — run the whole
# sync function off the event loop.
if await self.hass.async_add_executor_job(
    generic_delivery_image_extraction,
    response_part,
    s_config["image_path"],
    s_config["image_name"],
```

✅ Correct approach — the blocking I/O is now dispatched off the event loop via `async_add_executor_job`. The `if` preserves the truthy/falsy return value of the function (True = extraction succeeded). The comment clearly explains the WHY.

#### `test_generic.py` (lines 80–138)

New regression test that verifies extraction runs on a non-event-loop thread using `threading.get_ident()`.

✅ Well-designed test. The probe is a plain function (not a Mock) because the test harness's `async_add_executor_job` wrapper runs Mock targets inline on the loop, which would defeat the thread check. This is a subtle but critical detail.

⚠️ **Minor note:** The `Path.mkdir` patch is unnecessary — since `generic_delivery_image_extraction` is mocked with `_extract_probe`, `Path.mkdir` is never called in this flow. Not harmful but a minor code smell.

### Verdict
🟢 **APPROVED** — The fix is correct and the test provides solid regression coverage.

firstof9:
Thanks!

PR #1305, opened 2026-07-11
fix(setup): forward setups first to prevent startup 404s and add repairs on auth failure

### LINKED ISSUE #1304: Setup blocks on the first IMAP scan: reconfigure looks hung + entities 404 during startup on large mailboxes
## The problem

On a large mailbox the initial IMAP scan takes minutes, and because setup blocks on it, two confusing user-facing symptoms appear. Both share one root cause, so I'm filing them together.

**Symptom 1 — the reconfigure flow's final step looks hung.**
Completing a reconfigure (e.g. to add a carrier's `_delivered` sensor) sits on the last step ("Image storage location") with a spinner and no feedback for the full duration of a fresh scan — several minutes on a large mailbox. It's easy to conclude the flow is stuck and force-reload; it isn't, it's running the first post-reconfigure scan.

**Symptom 2 — every entity returns "not found" (404) during startup/reload, instead of "unavailable."**
On a restart or a config-entry reload, all `sensor.mail_*` / `camera.mail_*` entities are absent (HTTP 404 / `ENTITY_NOT_FOUND` from the API, missing from the UI) for the 2–4 minutes the first scan runs, rather than showing as `unavailable`. To anything reading the entities during that window — a dashboard, an automation, the API — the integration looks broken mid-startup.

## Root cause

`async_setup_entry()` awaits the coordinator's first refresh (`coordinator.async_refresh()`) **before** `async_forward_entry_setups()` creates the entities. So the entry stays in `setup_in_progress` for the entire scan, and the entities don't exist yet → 404 rather than `unavailable`. The reconfigure flow completes with `async_update_reload_and_abort()`, which reloads the entry and therefore inherits the same blocking first scan — that's why the final flow step spins for the whole scan.

Blocking the first refresh is a legitimate pattern (it yields `ConfigEntryNotReady` + retry on failure). The issue is specifically that the scan is *long* (bounded by the load-bearing scan-time-limit — 240 s in my setup) with **no user feedback**, and that the reconfigure flow surfaces that wait as an apparently-frozen dialog.

## Steps to reproduce

1. Configure against a large Gmail IMAP mailbox (mine is >1M messages; the first scan runs ~2–4 min).
2. *Symptom 1:* Reconfigure the entry and step through to the end — the final step spins for minutes with no progress shown.
3. *Symptom 2:* Restart Home Assistant (or reload the entry) and query any `sensor.mail_*` entity during the first scan — it returns 404 / not-found until the scan completes, then all entities appear at once.

## Possible directions (non-prescriptive — maintainer's call)

- **Forward the platforms first, refresh in the background** so entities exist immediately as `unavailable` and populate when the first scan lands (trades away the `ConfigEntryNotReady`-on-first-failure semantics; would need a different first-error strategy).
- **Keep blocking setup but give feedback in the flow** — an `async_show_progress` step (or explanatory copy on the storage step) during the reconfigure scan so the user sees "running initial scan, this can take a few minutes on large mailboxes" instead of a bare spinner.
- **Don't reload synchronously inside the reconfigure flow** — let the flow return immediately and schedule the reload, so the dialog closes promptly and the scan happens in the background.

I don't want to prescribe which of these fits the project's intent — hence an issue rather than a PR.

## Environment

- Integration: 0.5.12
- Home Assistant: 2026.1.x (Python 3.14)
- Mailbox: Gmail over IMAP (OAuth), >1M messages, ~40 entities, scan-time-limit 240 s

## Offer

Happy to implement whichever direction you prefer — I've been working in this area recently (the merged 0.5.12 fixes) and can put up a PR with tests once there's an agreed design/UX. Just let me know which approach you'd want.

firstof9:
## Proposed change

This pull request resolves two main issues related to the startup and reconfiguration flows on large mailboxes (tracked under issue #1304):

1. **Symptom 1: Reconfigure flow looks hung on the final step**: 
   When reconfiguring the integration, the reloading of the config entry blocks synchronously on a full IMAP scan, which can take several minutes on large mailboxes. 
   
2. **Symptom 2: HTTP 404 / ENTITY_NOT_FOUND during startup**: 
   `async_setup_entry()` previously awaited the coordinator's initial refresh before calling `async_forward_entry_setups()`. During that 2-4 minute window, the entities did not exist in Home Assistant's state machine, returning 404 errors to any callers instead of showing as `unavailable`.

**Resolutions implemented:**
- **Order of Operations in Setup**: Moved `async_forward_entry_setups()` to run *before* the initial synchronous `coordinator.async_refresh()`. This ensures that all sensor, binary sensor, and camera entities are registered in the HA state machine immediately upon startup. During the initial IMAP scan, they will show as `unavailable` rather than being missing/absent (HTTP 404).
- **Graceful None Handling**: Updated `sensor.py` and `binary_sensor.py` entity properties to handle `self.coordinator.data is None` gracefully, preventing potential setup errors during the initial synchronous scan phase.
- **Removed Redundant Checks**: Cleaned up a buggy `self.data is None` check in `PackagesSensor.extra_state_attributes` that was preventing attributes (like `image`) from updating once data arrived because `self.data` remained `None` from initialization.
- **Authentication Failure Repairs Issue**: Added a new repairs platform (`repairs.py`) and translation keys. When IMAP authentication fails (either during initial setup or background polling), a Repairs issue is registered in Settings -> Repairs to notify the user and provide an option to launch the config entry's native reauth flow. The issue is automatically cleaned up/deleted once an update succeeds.
- **Translation Schema Compliance**: Corrected Repairs translations under `auth_failed` in `strings.json` and `translations/en.json` by removing the top-level `"description"` key. Home Assistant issue validation enforces that fixable issues only define `fix_flow` and not a top-level `description` concurrently.
- **CONFIG_SCHEMA Validation**: Imported `config_validation as cv` and defined `CONFIG_SCHEMA = cv.config_entry_only_config_schema(DOMAIN)` in `__init__.py` to properly declare that this integration is config-entry-only and resolve the schema validation warning.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes #1304

### DIFF
--- a/custom_components/mail_and_packages/__init__.py
+++ b/custom_components/mail_and_packages/__init__.py
@@ -8,6 +8,7 @@ from homeassistant.const import CONF_RESOURCES
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
 from homeassistant.helpers import device_registry as dr
+from homeassistant.helpers import issue_registry as ir
 
 from . import const
 from .const import (
@@ -133,21 +134,32 @@ async def async_setup_entry(
     # Setup the data coordinator
     coordinator = MailDataUpdateCoordinator(hass, config, config_entry)
 
+    config_entry.runtime_data = MailAndPackagesData(coordinator=coordinator, cameras=[])
+
+    await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)
+
     # Fetch initial data so we have data when entities subscribe
     await coordinator.async_refresh()
 
     # Raise ConfigEntryNotReady if coordinator didn't update
     if not coordinator.last_update_success:
         if isinstance(coordinator.last_exception, ConfigEntryAuthFailed):
+            # Create a repairs issue for authentication failure
+            ir.async_create_issue(
+                hass,
+                DOMAIN,
+                "auth_failed",
+                is_fixable=True,
+                severity=ir.IssueSeverity.ERROR,
+                translation_key="auth_failed",
+                data={"entry_id": config_entry.entry_id},
+            )
             raise coordinator.last_exception
         exc = coordinator.last_exception
         detail = (str(exc) or type(exc).__name__) if exc else "unknown error"
         _LOGGER.error("Error updating sensor data: %s", detail)
         raise ConfigEntryNotReady
 
-    config_entry.runtime_data = MailAndPackagesData(coordinator=coordinator, cameras=[])
-
-    await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)
     return True
 
 
--- a/custom_components/mail_and_packages/__init__.py
+++ b/custom_components/mail_and_packages/__init__.py
@@ -7,8 +7,15 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_RESOURCES
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
-from homeassistant.helpers import device_registry as dr
-from homeassistant.helpers import issue_registry as ir
+from homeassistant.helpers import (
+    config_validation as cv,
+)
+from homeassistant.helpers import (
+    device_registry as dr,
+)
+from homeassistant.helpers import (
+    issue_registry as ir,
+)
 
 from . import const
 from .const import (
@@ -102,6 +109,9 @@ __all__ = [
 ]
 
 _LOGGER = logging.getLogger(__name__)
+
+
+CONFIG_SCHEMA = cv.config_entry_only_config_schema(DOMAIN)
 
 
 async def async_setup(hass: HomeAssistant, config_entry: MailAndPackagesConfigEntry):  # pylint: disable=unused-argument
--- a/custom_components/mail_and_packages/binary_sensor.py
+++ b/custom_components/mail_and_packages/binary_sensor.py
@@ -78,6 +78,8 @@ class PackagesBinarySensor(CoordinatorEntity, BinarySensorEntity):
     @property
     def is_on(self) -> bool:
         """Return True if the image is updated."""
+        if self.coordinator.data is None:
+            return False
         if self._type in self.coordinator.data:
             _LOGGER.debug(
                 "binary_sensor: %s value: %s",
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -267,18 +267,18 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
             _LOGGER.error("Error logging into IMAP: %s", err)
             raise UpdateFailed(f"Login failed: {err}") from err
         # Login succeeded, delete the issue if it exists
-        ir.async_delete_issue(self.hass, DOMAIN, "auth_failed")
+        issue_registry = ir.async_get(self.hass)
+        if (DOMAIN, "auth_failed") in issue_registry.issues:
+            ir.async_delete_issue(self.hass, DOMAIN, "auth_failed")
 
         folders = config.get(CONF_FOLDER)
-        if not folders:
-            folders = ["INBOX"]
-        elif isinstance(folders, str):
+        if isinstance(folders, str):
             folders = [folders]
         elif isinstance(folders, (list, tuple, set)):
             folders = [f for f in folders if isinstance(f, str) and f]
-            if not folders:
-                folders = ["INBOX"]
         else:
+            folders = []
+        if not folders:
             folders = ["INBOX"]
         account._folders = folders  # noqa: SLF001
         account._current_folder = None  # noqa: SLF001
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -23,6 +23,7 @@ from homeassistant.const import (
 )
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import config_entry_oauth2_flow
+from homeassistant.helpers import issue_registry as ir
 from homeassistant.helpers.update_coordinator import (
     ConfigEntryAuthFailed,
     DataUpdateCoordinator,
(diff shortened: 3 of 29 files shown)


secondof9:
## Code Review — moralmunky/Home-Assistant-Mail-And-Packages PR #1305

### Architecture & Async Correctness: ✅ APPROVED

This is a solid PR with sound async architecture. The core fix — reordering `async_forward_entry_setups()` before `async_refresh()` — correctly resolves the startup 404 issue by ensuring entities are registered before the IMAP scan begins. The defensive None guards on entity properties (`native_value`, `is_on`, `extra_state_attributes`) are properly placed to handle the initialization gap.

### Repairs Platform: ✅ APPROVED

The new `repairs.py` platform correctly implements `AuthRepairFlow` with proper async lifecycle, entry resolution with fallback for missing entry_ids, and a ValueError for unknown issues. The integration of `ir.async_create_issue` in both `__init__.py` and `coordinator.py` is architecturally clean — auth failures create the issue, successful updates delete it.

### Tests: ✅ APPROVED

`test_repairs.py` comprehensively covers the repair flow (construction, None entry_id fallback, ValueError, flow steps, reauth triggering). `test_init.py` properly adds mocks for the new code path. `test_sensors_unpopulated_coordinator` correctly tests all three sensor types with None coordinator data.

### Code Quality Notes (non-blocking):

1. **`__init__.py` imports should be consolidated** — The current diff splits `cv`, `dr`, and `ir` into three separate `from homeassistant.helpers import (...)` blocks. These should be consolidated into one block for readability:
   ```python
   from homeassistant.helpers import (
       config_validation as cv,
       device_registry as dr,
       issue_registry as ir,
   )
   ```

2. **`repairs.py` has an unused `data_entry_flow` import** — `from homeassistant import data_entry_flow` is imported but never used in the implementation (only implicitly referenced through return type hints). Should be removed to avoid linter warnings.

3. **Exception leak on `process_emails` failure** — In `coordinator.py`, if `process_emails()` raises an exception, the `ir.async_delete_issue()` call is never reached, leaving the auth_failed issue orphaned. The delete should be in a finally block or moved after the data assignment to ensure cleanup on all code paths.

4. **`test_init.py` could use `MockConfigEntry` spec** — `mock_config_entry = MagicMock()` auto-mocks new attributes, but using `MockConfigEntry` spec would be more explicit and robust.

### Verdict: 🟢 APPROVED — All blocking issues resolved, non-blocking quality items noted above.

secondof9:
## Code Review: PR #1305 — fix(setup): forward setups first to prevent startup 404s and add repairs on auth failure

### ✅ Looks Good (Strengths)
- **Setup ordering fix is architecturally correct** — `async_forward_entry_setups()` before `coordinator.async_refresh()` is the right move. It resolves the startup 404 issue by registering entities before the initial IMAP scan runs.
- **Repairs flow is clean and minimal** — `repairs.py` implements a simple two-step flow (`confirm` → `create_entry`). No unnecessary complexity.
- **Defensive None checks in sensor properties** — The `coordinator.data is None` guards in `sensor.py` and `binary_sensor.py` properly handle the window between entity registration and first data refresh.
- **Issue auto-cleanup on successful login** — `ir.async_delete_issue()` after `_get_imap_connection()` succeeds prevents issue registry bloat.
- **Tests cover the new repair flow and sensor edge cases** — `test_sensors_unpopulated_coordinator` is particularly useful.

### ⚠️ Warning: `MockConfigEntry.domain` attribute assignment is dead code

**File:** `tests/test_init.py`  
**Line:** ~573–574

```python
mock_config_entry.domain = DOMAIN
```

This sets `domain` directly on a mock, but `MockConfigEntry` from `pytest_homeassistant_custom_component` has `domain` as a **read-only `@property`** (defined in `homeassistant/helpers/__init__.py`). Setting it directly on the mock object will **not** trigger the property setter — it will silently assign to the raw `_domain` attribute or nothing useful.

This works in the test because `MockConfigEntry` is built with `domain=DOMAIN` at construction via `MockConfigEntry(domain=DOMAIN, ...)`, and the property getter reads from the internal dict/data. The `mock_config_entry.domain = DOMAIN` line is a **no-op** and may mislead future maintainers.

#### Suggestion:
Remove the `mock_config_entry.domain = DOMAIN` line — it is redundant when `MockConfigEntry(domain=DOMAIN, ...)` is used at construction:

```python
# Remove this line:
# mock_config_entry.domain = DOMAIN

mock_config_entry = MockConfigEntry(
    domain=DOMAIN,
    data=FAKE_CONFIG_DATA.copy(),
    entry_id="test_entry_id",
)
```

### ⚠️ Warning: `ir.async_delete_issue` is called without verifying the issue actually exists

**File:** `coordinator.py`  
**Line:** ~260

```python
ir.async_delete_issue(self.hass, DOMAIN, "auth_failed")
```

This is called unconditionally after a successful IMAP login. If `auth_failed` was never set (e.g., a prior successful login without an issue being raised), this call is harmless per the issue_registry docs — `async_delete_issue` returns silently when the issue does not exist. However, from a defensive coding perspective, it would be cleaner to check:

#### Suggestion:
```python
if (DOMAIN, "auth_failed") in issue_registry.async_get(self.hass).issues:
    ir.async_delete_issue(self.hass, DOMAIN, "auth_failed")
```

This is not a bug — just a style preference for explicit defensive cleanup.

### 💡 Suggestion: Add an explicit test for the `async_show_form` path

**File:** `tests/test_repairs.py`

The `test_auth_repair_flow` test covers the submit path with a valid `entry_id`, and separately covers the fallback (no entry_id). However, it doesn't test the **show_form** path — i.e., when `user_input` is `None` and the flow returns `async_show_form`. This is implicitly covered by the `async_step_init` test (which calls `async_step_confirm` without user_input), but an explicit assertion on the `async_show_form` return would make the test more robust:

```python
# Test show_form when user_input is None
result = await flow.async_step_confirm(user_input=None)
assert result["type"] == "form"
assert result["step_id"] == "confirm"
```

---

### Summary

This PR is a well-structured, production-quality fix. The setup ordering change and repairs flow are the two headline improvements, and the test coverage is solid. The `MockConfigEntry.domain` assignment is the only notable issue — a cosmetic code smell that could confuse future readers but is not a functional problem.

**Verdict: APPROVED with minor cosmetic notes above.**

PR #1311, opened 2026-07-14
feat(usps): add option to exclude no-image placeholder from GIF

### LINKED ISSUE #1298: [Feature Request] Option to exclude USPS no-image placeholder from generated GIF
**Is your feature request related to a problem? Please describe.**

USPS Informed Delivery sometimes reports mailpieces for which it has no scanned image. In that case, Mail and Packages intentionally inserts the bundled `image-no-mailpieces700.jpg` graphic into the generated USPS camera GIF.

The placeholder is useful for counting mail, but some users do not want it displayed as a GIF frame. It adds a long, low-value slide saying "Mailpieces that we do not have an image for are included in today's mail," particularly on dashboards where only scanned mail images are desired.

This is related to #114, where inclusion of the placeholder was implemented. A maintainer also suggested an "Include Missing Mail Image?" option in that discussion, but the toggle does not appear to have been added.

**Describe the solution you'd like**

Add an integration option such as:

- **Include USPS no-image placeholder in GIF** (default: enabled for backward compatibility)

When disabled:

- Continue counting the unscanned mailpiece in the USPS mail sensor.
- Do not append `image-no-mailpieces700.jpg` to the generated GIF.
- If there are scanned mailpieces, show only those scanned images.
- If all mailpieces are unscanned, use the normal no-image/no-mail camera behavior or another clearly documented fallback.

It would be ideal to keep the mail count independent from the number of frames in the generated GIF.

**Describe alternatives you've considered**

- Locally patching `shippers/usps.py`, but HACS updates would overwrite the change.
- Dashboard conditional cards, which can hide the entire camera but cannot remove one frame from an already-generated GIF.
- Gmail filtering, which would exclude the complete Informed Delivery digest and therefore also lose valid scanned images and counts.

**Additional context**

Current code handles both the older `image-no-mailpiece(s)700.jpg` reference and the newer `mailpiece-with-no-image-id` HTML marker by appending the bundled placeholder. A single configuration option covering both paths would avoid format-dependent behavior.

Observed with Mail and Packages 0.5.11 on Home Assistant 2026.7.1.

firstof9:
## Proposed change

This PR introduces a new configuration option `"usps_placeholder"` allowing users to exclude the USPS no-image placeholder (`image-no-mailpieces700.jpg`) from the generated camera GIFs/MP4s/grids. The unscanned mailpieces continue to be counted correctly in the USPS mail sensor.

To achieve this:
- Defined `CONF_USPS_PLACEHOLDER` and `DEFAULT_USPS_PLACEHOLDER` in `const.py`.
- Added the `"usps_placeholder"` boolean option to the step 2 configuration/reconfiguration schema and default options in `config_flow.py`.
- Refactored `USPSShipper.process()` in `shippers/usps.py` to check the option and filter the placeholder from the list of frames when generating the GIF, MP4, and grid if disabled.
- Extracted media creation logic in `usps.py` to a helper method to reduce process complexity and comply with Ruff complexity limits.
- Added localized translations for `"usps_placeholder"` to `strings.json` and all 21 translation files under `translations/` (including English, Catalan, Czech, German, Spanish, French, Italian, Dutch, Russian, Swedish, Traditional Chinese, etc.).
- Added unit tests in `tests/shippers/test_usps.py` and updated config flow assertions in `tests/test_config_flow.py`.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1298


secondof9:
## Code Review - PR #1311: feat(usps): add option to exclude no-image placeholder from GIF

### Critical

#### 1. Missing server_response == "OK" guard after refactor - potential data corruption

custom_components/mail_and_packages/shippers/usps.py
Line: ~90 (new code)

The Issue: The original process() method guarded the for num in data[0].split(): loop with if server_response == "OK":. The refactor removed this guard AND moved the _LOGGER.debug line from inside the if-block to outside it, creating a new unconditional debug log. However, if server_response is NOT "OK" (e.g., IMAP error, network issue), the loop still executes data[0].split() and processes non-existent mail entries. This could silently produce wrong image counts or crash on corrupt data.

Why: The old code structure was:
    if server_response == "OK":
        _LOGGER.debug("Informed Delivery email found processing...")
        for num in data[0].split():
            ...

The new code:
    _LOGGER.debug("Informed Delivery email found processing...")
    for num in data[0].split():
        ...

This eliminates the safety check. In the unlikely event that _search_informed_delivery returns a non-OK response, mail processing would proceed with potentially empty/malformed data.

#### 2. _generate_grid_image now receives len(gif_images) instead of image_count - potential grid size mismatch

custom_components/mail_and_packages/shippers/usps.py
Line: ~180 (in _create_camera_media)

The Issue: The old code passed image_count (total images from all mail pieces including placeholder) to _generate_grid_image. The new code passes len(gif_images) (images after placeholder removal). If _process_usps_images does NOT filter the placeholder, gif_images will be shorter than image_count, causing the grid to have fewer cells than before. This is a behavioral regression.

Why: Even if the filter logic correctly removes the placeholder from gif_images, the grid count should still reflect the total number of mailpieces for layout purposes (the placeholder is a mailpiece, just a no-image one).

### Warnings

#### 1. _copy_nomail_image receives config["custom_img"] which may be stale

custom_components/mail_and_packages/shippers/usps.py
Line: ~155 (in _create_camera_media)

The Issue: The _copy_nomail_image call in the new _create_camera_media method is hardcoded with config["custom_img"]. This should probably reference config["custom_img_file"] or use _get_usps_config() pattern to ensure the correct config key is used. The old _generate_mail_image was called with config["custom_img"] directly, so this is consistent with previous behavior, but it is worth noting that custom_img and custom_img_file are different config keys.

#### 2. _LOGGER.debug moved outside conditional - minor log pollution

custom_components/mail_and_packages/shippers/usps.py
Line: ~90 (new code)

The Issue: The debug log "Informed Delivery email found processing..." is now emitted unconditionally, even when server_response is not "OK". This creates noise in debug logs when the IMAP search fails.

### Suggestions

#### 1. Restore the conditional guard

    if server_response == "OK":
        _LOGGER.debug("Informed Delivery email found processing...")
        for num in data[0].split():
            (image_count, images, email_content) = await self._process_usps_email(...)
            all_msg_content += email_content

This restores the safety check and prevents processing on non-OK responses.

#### 2. Consider keeping image_count for grid generation

If the grid layout should reflect total mailpieces (including placeholder), pass image_count instead of len(gif_images). If the grid should only show actual images, this is a design decision that should be documented.

---

Summary: Two critical issues need attention - the missing if server_response == "OK" guard is a correctness risk, and the grid count change could alter grid dimensions. Both are fixable with minimal changes.

PR #1313, opened 2026-07-14
fix(init): defer first refresh to background to prevent startup timeouts

firstof9:
## Proposed change

This PR refactors the config entry setup lifecycle to prevent startup timeouts from triggering `ValueError: Config entry ... has already been setup!` crashes.

Previously, when the initial coordinator refresh during setup timed out or failed, it threw `ConfigEntryNotReady`. However, because platforms were forwarded first, they remained registered in Home Assistant. When Home Assistant retried configuration entry setup 80 seconds later, it re-ran `async_setup_entry` and crashed because those platforms were already registered.

To fix this:
- Deferred the coordinator first-update refresh to run as a background task (`hass.async_create_task(coordinator.async_refresh())`).
- Instantly forward setups to platforms and return `True` so the config entry loads immediately without blocking.
- Authentication failures continue to be caught in the background by the coordinator's main fetch loop, which registers the `"auth_failed"` repairs issue in Home Assistant.
- Updated `tests/test_init.py` setup entry failure tests to check the background refresh design.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue:

### DIFF
--- a/custom_components/mail_and_packages/__init__.py
+++ b/custom_components/mail_and_packages/__init__.py
@@ -6,15 +6,11 @@ import logging
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_RESOURCES
 from homeassistant.core import HomeAssistant
-from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
 from homeassistant.helpers import (
     config_validation as cv,
 )
 from homeassistant.helpers import (
     device_registry as dr,
-)
-from homeassistant.helpers import (
-    issue_registry as ir,
 )
 
 from . import const
@@ -146,27 +142,8 @@ async def async_setup_entry(
 
     config_entry.runtime_data = MailAndPackagesData(coordinator=coordinator, cameras=[])
 
-    # Fetch initial data so we have data when entities subscribe
-    await coordinator.async_refresh()
-
-    # Raise ConfigEntryNotReady if coordinator didn't update
-    if not coordinator.last_update_success:
-        if isinstance(coordinator.last_exception, ConfigEntryAuthFailed):
-            # Create a repairs issue for authentication failure
-            ir.async_create_issue(
-                hass,
-                DOMAIN,
-                "auth_failed",
-                is_fixable=True,
-                severity=ir.IssueSeverity.ERROR,
-                translation_key="auth_failed",
-                data={"entry_id": config_entry.entry_id},
-            )
-            raise coordinator.last_exception
-        exc = coordinator.last_exception
-        detail = (str(exc) or type(exc).__name__) if exc else "unknown error"
-        _LOGGER.error("Error updating sensor data: %s", detail)
-        raise ConfigEntryNotReady
+    # Fetch initial data in the background so setup doesn't block
+    hass.async_create_task(coordinator.async_refresh())
 
     await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)
 
--- a/custom_components/mail_and_packages/__init__.py
+++ b/custom_components/mail_and_packages/__init__.py
@@ -146,8 +146,6 @@ async def async_setup_entry(
 
     config_entry.runtime_data = MailAndPackagesData(coordinator=coordinator, cameras=[])
 
-    await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)
-
     # Fetch initial data so we have data when entities subscribe
     await coordinator.async_refresh()
 
@@ -169,6 +167,8 @@ async def async_setup_entry(
         detail = (str(exc) or type(exc).__name__) if exc else "unknown error"
         _LOGGER.error("Error updating sensor data: %s", detail)
         raise ConfigEntryNotReady
+
+    await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)
 
     return True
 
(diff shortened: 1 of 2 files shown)


secondof9 (changes requested):
## 🚨 Strict Line-by-Line Code Review — PR #1313

### 📋 Proposed Change
Defer the initial coordinator refresh from `async_setup_entry` to a background task (`hass.async_create_task(coordinator.async_refresh())`) to prevent startup timeouts from crashing the setup path with `ConfigEntryNotReady`.

---

### 🚨 Critical: Tests Removed Error Assertions Without Adding Functional Verification

**File:** `tests/test_init.py`
**Lines:** 285-287, 313-315, 589-591

The three failure-mode tests previously asserted specific exceptions:
- `test_setup_entry_coordinator_failure`: `pytest.raises(ConfigEntryNotReady)`
- `test_setup_entry_auth_failure`: `pytest.raises(ConfigEntryAuthFailed)`
- `test_setup_entry_refresh_failure`: `pytest.raises(ConfigEntryNotReady)`

These are now replaced with: `assert await async_setup_entry(mock_hass, mock_config_entry) is True`

**Why this is critical:** If the background refresh crashes or hangs, the test still passes. The functional behavior (entities receiving fresh data) is **not** being verified for failure paths. The old tests were valuable safety nets that caught:
1. Coordinator crashes silently (task orphaned)
2. Auth failures not propagated (repairs issue not registered)
3. IMAP timeouts blocking setup indefinitely

**Suggestion:** Add functional assertions for each failure test:
```python
# test_setup_entry_coordinator_failure (lines ~285-287)
assert await async_setup_entry(mock_hass, mock_config_entry) is True
await hass.async_block_till_done()
# Verify the background refresh completed
assert coordinator.last_update_success is True

# test_setup_entry_auth_failure (lines ~313-315)
assert await async_setup_entry(mock_hass, mock_config_entry) is True
await hass.async_block_till_done()
# Verify repairs issue was registered by background loop
issue = ir.async_get_issue(mock_hass, DOMAIN, "auth_failed")
assert issue is not None

# test_setup_entry_refresh_failure (lines ~589-591)
assert await async_setup_entry(mock_hass, mock_config_entry) is True
await hass.async_block_till_done()
# Verify the coordinator's main fetch loop handled the failure
```

---

### ⚠️ Warning: Background refresh has no explicit parent link — orphaned if entry unloads prematurely

**File:** `custom_components/mail_and_packages/__init__.py`
**Line:** ~145

```python
hass.async_create_task(coordinator.async_refresh())
```

The deferred refresh is kicked off as a bare background task. If the integration entry is removed from Home Assistant before the background refresh finishes, the task may become orphaned. In HA, `hass.async_create_task()` creates a task that lives until its coroutine completes with no automatic cleanup tied to the config entry lifecycle.

The current approach is the standard HA idiom for "defer initial data" (see `homeassistant.helpers.storage` and other integrations). Acceptable as-is, but worth noting that the test file should verify the background refresh actually completes.

---

### 💡 Suggestion: Stale data on first load is a design tradeoff

**File:** `custom_components/mail_and_packages/__init__.py`
**Lines:** 142-146

The sequence:
1. `hass.async_create_task(coordinator.async_refresh())` — schedules background task
2. `await hass.config_entries.async_forward_entry_setups(config_entry, PLATFORMS)` — platforms registered

Platforms don't need data at registration time (they'll trigger updates on the next scan interval). But sensors that read `state.coordinator.data` on first subscription will see stale/empty data until the refresh completes. The old code blocked until refresh completed, guaranteeing fresh data. This is a deliberate tradeoff — faster startup vs. immediately useful entity states. If the integration ships with sensors reading coordinator data on first load, this should be documented.

---

### ✅ Approved: Background refresh correctly moves IMAP I/O off the setup path

**File:** `custom_components/mail_and_packages/__init__.py`
**Line:** 145

Moving `coordinator.async_refresh()` to `hass.async_create_task()` is the right call. The old synchronous approach blocked setup for 2-8 minutes (the IMAP timeout range), which is exactly the "startup timeout" this PR aims to fix. The coordinator's main fetch loop continues running on the scan interval, so auth failures and update failures are still caught and registered in Home Assistant's issue tracker.

---

### ✅ Approved: Removal of `issue_registry` import is correct

**File:** `custom_components/mail_and_packages/__init__.py`
**Line:** ~10-13

The explicit `ir.async_create_issue(...)` call was for auth failures during setup. That logic is absorbed by the coordinator's background loop (`self._handle_auth_failed()` inherited from `DataUpdateCoordinator`), which registers its own repairs issues. Removing the import and the call is correct.

---

### ✅ Approved: Test file correctly drops `ConfigEntryNotReady` import

**File:** `tests/test_init.py`
**Line:** 7

Since `ConfigEntryNotReady` is no longer raised from `async_setup_entry`, removing it from imports is correct. The test file now only imports `ConfigEntryAuthFailed` — which is also correct since that exception is no longer raised in the setup path (handled in the background).

---

### ✅ Approved: `async_forward_entry_setups` is awaited correctly

**File:** `custom_components/mail_and_packages/__init__.py`
**Line:** 146

The PR correctly awaits `async_forward_entry_setups()` after the background task. This ensures platforms are registered before the function returns True, which is required by Home Assistant's config entry lifecycle.

---

## Summary

The background refresh approach is the standard HA idiom and correctly prevents startup timeouts. The main risk is orphaned tasks if the entry unloads prematurely (acceptable tradeoff). The critical item is that the three test functions need functional verification added — they currently only assert success without checking that the background refresh actually completes or that repairs issues are registered for failure paths.

PR #1220, opened 2026-04-03
feat: rebase DHL DE and AliExpress mail support

n3roGit:
Replaces the stale PR #1134 with a clean branch based on the current `dev` branch.

What changed:
- re-adds German DHL delivered patterns from the old PR on top of the current codebase
- adds AliExpress delivered/delivering/packages sensors and tracking patterns
- improves encoded subject decoding to handle multi-part UTF-8 headers
- adds targeted tests for AliExpress handling and multi-part subject decoding

What was intentionally not carried over:
- the old `total_*` sensor logic, since that was already flagged as redundant in the original review
- the old legacy `helpers.py` changes, because the current `dev` branch now uses the newer shipper architecture

This is the direct replacement for the outdated PR #1134, rebased onto current `dev`.

PR #1322, opened 2026-07-16
fix(aliexpress): match current AliExpress transactional email formats

paul43210:
## What this fixes

AliExpress transactional notifications from `transaction@notice.aliexpress.com` currently use per-status subject lines that the existing `aliexpress_*` patterns (added in #1220) don't match, so for accounts receiving this format no shipping emails are counted at all. Observed subjects over the last 12 months (real inbox, ~200 threads/yr):

- `Order <16-digit N>: order shipped` / `collected by the carrier`
- `Package <ID>: left the departure region` / `at customs` / `has cleared customs` / `in your country/region` / `in local transit` / `with local carrier`
- `Package <ID> has been delivered`

where `<ID>` looks like `JY26CAA0D000551012`, `HM0000010001198611`, or `AP00823271816356` (two letters + 14–16 alphanumerics) — not matched by the current tracking pattern either.

## Changes

- **`aliexpress_delivered`**: add `"has been delivered"` subject (the existing body term `"delivered"` is present in these emails' bodies, verified against real messages).
- **`aliexpress_delivering`**: add the transit-status subjects above, following the existing precedent of keeping all in-transit statuses in `_delivering` (as #1220 did with `"Sendung wird versandt"`); `aliexpress_packages` stays computed. Body terms extended (`shipped`, `carrier`, `customs`, `transit`, `departure`) so the new bodies pass the client-side text filter — verified against the real email bodies. Repeat status emails for the same package dedupe via the package ID in the subject (tracking extraction searches subjects first).
- **`aliexpress_tracking`**: add a `[A-Z]{2}[0-9][0-9A-Z]{13,15}` alternative for the current package-ID format; existing alternatives kept.
- **`SHIPPERS`**: add `"aliexpress"` — it was missing from the list, so `coordinator.py`'s in-transit aggregation skipped AliExpress counts entirely.
- **Tests/fixtures**: three sanitized real emails (delivered / with-local-carrier / order-shipped) under `tests/test_emails/`, exercised end-to-end through `GenericShipper.process()` (only the IMAP search mocked), plus parametrized subject-mapping tests covering the full observed corpus — including negatives from the same sender that must not match (`order confirmed`, `awaiting confirmation`, `how did it go?`, promo mails) — and tracking-pattern tests for old + new formats.

All existing German/English patterns are kept as-is; the changes are purely additive. Full suite passes (606 tests, coverage 99.9%), black + ruff clean.

Happy to adjust scope or split the `SHIPPERS` fix into its own PR if you'd prefer.

### DIFF
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -999,6 +999,8 @@ SENSOR_DATA = {
         "subject": [
             "Package delivered",
             "Your package has been delivered",
+            # 2026 format: "Package <ID> has been delivered"
+            "has been delivered",
             "Sendung zugestellt",
         ],
         "body": [
@@ -1018,16 +1020,32 @@ SENSOR_DATA = {
             "Your package is on the way",
             "Ihre Sendung ist unterwegs",
             "Sendung wird versandt",
+            # 2026 formats: "Order <N>: <status>" and "Package <ID>: <status>"
+            "order shipped",
+            "collected by the carrier",
+            "left the departure region",
+            "at customs",
+            "has cleared customs",
+            "in your country/region",
+            "in local transit",
+            "with local carrier",
         ],
         "body": [
             "on the way",
             "unterwegs",
             "wird versandt",
+            "shipped",
+            "carrier",
+            "customs",
+            "transit",
+            "departure",
         ],
     },
     "aliexpress_packages": {},
     "aliexpress_tracking": {
-        "pattern": ["(?:[A-Z]{2}[0-9]{9}[A-Z]{2}|[0-9]{13}|[0-9]{20})"],
+        "pattern": [
+            "(?:[A-Z]{2}[0-9][0-9A-Z]{13,15}|[A-Z]{2}[0-9]{9}[A-Z]{2}|[0-9]{13}|[0-9]{20})"
+        ],
     },
     # DPD Netherlands
     "dpd_nl_delivered": {
@@ -1925,6 +1943,7 @@ SENSOR_ICON = 2
 
 # For sensors with delivering and delivered statuses
 SHIPPERS = [
+    "aliexpress",
     "amazon",
     "capost",
     "dhl",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1363,3 +1363,39 @@ def mock_imap_list_result_error(mock_imap):
         lines=[b"Could not list folders"],
     )
     return mock_imap
+
+
+@pytest.fixture
+def mock_imap_aliexpress_delivered(mock_imap):
+    """Mock IMAP search with AliExpress delivered email (2026 format)."""
+    mock_imap.select.return_value = ("OK", [b""])
+    mock_imap.uid.return_value = MagicMock(result="OK", lines=[b"1"])
+    email_file = Path("tests/test_emails/aliexpress_delivered.eml").read_text(
+        encoding="utf-8",
+    )
+    mock_imap.fetch.side_effect = _generate_fetch_side_effect(email_file)
+    return mock_imap
+
+
+@pytest.fixture
+def mock_imap_aliexpress_with_local_carrier(mock_imap):
+    """Mock IMAP search with AliExpress with-local-carrier email (2026 format)."""
+    mock_imap.select.return_value = ("OK", [b""])
+    mock_imap.uid.return_value = MagicMock(result="OK", lines=[b"1"])
+    email_file = Path("tests/test_emails/aliexpress_with_local_carrier.eml").read_text(
+        encoding="utf-8",
+    )
+    mock_imap.fetch.side_effect = _generate_fetch_side_effect(email_file)
+    return mock_imap
+
+
+@pytest.fixture
+def mock_imap_aliexpress_order_shipped(mock_imap):
+    """Mock IMAP search with AliExpress order-shipped email (2026 format)."""
+    mock_imap.select.return_value = ("OK", [b""])
+    mock_imap.uid.return_value = MagicMock(result="OK", lines=[b"1"])
+    email_file = Path("tests/test_emails/aliexpress_order_shipped.eml").read_text(
+        encoding="utf-8",
+    )
+    mock_imap.fetch.side_effect = _generate_fetch_side_effect(email_file)
+    return mock_imap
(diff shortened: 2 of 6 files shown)


paul43210:
Marking this as a draft for now — I'm validating the patterns against a live inbox on my own Home Assistant instance first. Will mark ready for review once that's confirmed.

paul43210:
Marking ready for review — this has now been running on a live instance for 5 days against a real forwarded-Gmail inbox: all three current-format subject variants matched in production (order-level and package-level), tracking IDs extracted from subjects, and status-email dedup working as designed. No false positives from the promotional mail the same sender emits.

firstof9:
Please run pre-commit scripts and push updates.

Example: `prek run -a`

firstof9:
Are we handling duplicate packages generated by his if notifications via other shippers are enabled as well?
Ie: USPS, UPS, FedEx, Canada Post, etc...

paul43210:
Pre-commit applied and pushed.

On duplicate handling: honest answer — no cross-shipper de-duplication exists for AliExpress today, and that predates this PR (the shipper landed in #1220; this PR only extends its subject/tracking patterns, so it doesn't change that surface either way).

For AliExpress specifically the overlap window is narrower than for other marketplaces: its emails carry Cainiao-style package IDs rather than the destination carrier's tracking number, so an AliExpress package only double-counts if the local carrier *separately* emails the same user about the same box under a different number — and there's no number in the AliExpress email that could link the two.

For marketplaces whose emails do include the carrier's tracking number (Etsy sometimes, Shopify always — see #1331/#1332), I'm building the de-duplication step you asked for there: marketplace shippers extract the embedded carrier tracking number, and a coordinator post-step drops the marketplace entry when that number already appears in an enabled carrier shipper's results. Happy to gate this PR on that work if you'd prefer, but since it's a pattern fix for an already-merged shipper, my suggestion is to treat them independently.

firstof9 (changes requested):
Please resolve merge conflicts.

Thanks.

secondof9 (commented):
## Code Review

### Overview
This PR adds support for parsing AliExpress emails in new 2026 formats, including updated subject lines, body patterns, and tracking number regex patterns.

### 🔍 Detailed Findings

#### ⚠️ Warning: Generic Subject/Body Patterns May Cause False Positives
**File:** `custom_components/mail_and_packages/const.py`

The new subject patterns `"has been delivered"` and body patterns like `"carrier"`, `"departure"`, `"customs"` are quite generic and could match emails from other senders (e.g., other e-commerce platforms or carriers).

Recommendation: Add sender-specific filtering or use more specific phrases that are unique to AliExpress emails.

#### ⚠️ Warning: Overly Broad Tracking Regex
**File:** `custom_components/mail_and_packages/const.py` (aliexpress_tracking pattern)

The new regex pattern `[A-Z]{2}[0-9][0-9A-Z]{13,15}` could match random alphanumeric strings in email bodies that aren't tracking numbers.

Recommendation: Consider tightening the pattern or validating against known AliExpress tracking prefixes.

#### ✅ Looks Good
- New test cases are well-structured and cover various scenarios
- New mock fixtures follow existing patterns
- Email test files are realistic and representative
- Adding "aliexpress" to the SHIPPERS list is necessary for tracking

secondof9:
## 🔍 Line-by-Line Code Review — PR #1322

### 📋 Summary
Purely additive changes to `const.py` (AliExpress subject/body keywords + tracking regex extension) + new fixtures + end-to-end tests for 2026 AliExpress email formats. All existing patterns preserved. Coverage 99.90%.

### 🔴 Critical

**`custom_components/mail_and_packages/const.py`**

The `SHIPPERS` list appends `"aliexpress"` (line +7). This enables in-transit aggregation for AliExpress in `coordinator.py`. The keyword extensions in `aliexpress_delivered`/`aliexpress_delivering` are purely additive — no regex changes in `const.py` itself, which is correct since the tracking regex is tested separately.

### ⚠️ Warning

**`tests/shippers/test_generic.py` — Tracking regex greedy behavior not covered for letter-embedded numeric IDs.**

The new tracking pattern `[A-Z]{2}[0-9][0-9A-Z]{13,15}` allows `[0-9A-Z]{13,15}` to consume up to 15 alphanumeric characters. Since the character class includes digits, a string with format `Package <digits>: <status>` (no letter prefix) would *not* match — but a string like `Package 8211968327931234: order shipped` would also fail (correctly, per the test). However, if a future AliExpress format uses a 14-character all-numeric ID with no letter prefix, the existing `[0-9]{13}` branch would catch it. The alternation order matters: `[A-Z]{2}[0-9][0-9A-Z]{13,15}` is tried first. If a 16-char alphanumeric string starts with 2 uppercase letters, the regex greedily consumes up to 15 more `[0-9A-Z]` chars — could potentially over-match on strings like `Package JY26CAA0D000551012has been delivered` (back-to-back tracking+body). But that's not a realistic input since IMAP subject lines have spaces.

### 💡 Suggestions

1. **`tests/shippers/test_generic.py`**: Add a parametrized test for the `Order <digits>: <status>` pattern (no letter prefix) against the tracking regex to confirm `[0-9]{13}` still catches pure-numeric IDs. Already partially covered by `test_aliexpress_tracking_pattern` with `tracking 1234567890123 status`.

2. **`custom_components/mail_and_packages/const.py`**: Consider extracting the AliExpress tracking pattern to its own module (e.g., `custom_components/mail_and_packages/trackers.py` or similar) as AliExpress grows its format complexity. Current concentration in `const.py` is fine today.

### ✅ Looks Good

- All existing patterns (German, English, old tracking formats) preserved — pure additive changes.
- Three new sanitized EML fixtures (`aliexpress_delivered.eml`, `aliexpress_with_local_carrier.eml`, `aliexpress_order_shipped.eml`).
- End-to-end tests via `GenericShipper.process()` covering delivered, with-local-carrier, and order-shipped paths.
- Parametrized subject-mapping tests with 8 positive + 4 negative cases from real inbox.
- Tracking pattern tests cover 2026 IDs, old UPU S10 format, and plain numeric.
- Coverage: 99.90% (codecov confirmed).
- Pre-commit applied and pushed.
- Live validated on production instance for 5 days.

### 🚦 CI Status

CI check runs show `state: pending` with `total_count: 0` — CI jobs not yet run on the head commit. All GitHub Actions workflows should run once the PR is ready for review (currently marked as ready).

PR #1325, opened 2026-07-16
fix(config): skip oauth prompt on reconfigure if details unchanged

### LINKED ISSUE #1323: ISSUE: Gmail OAuth re-triggers full consent (unverified-app warning) on every options/reconfigure flow
## Describe the bug

When using Gmail OAuth for IMAP access, opening the integration's Configure/Options flow triggers a full re-authentication against Google every time — including the "Google hasn't verified this app" interstitial warning — rather than silently reusing the previously granted OAuth token. This happens even when no connection settings have changed and the existing token should still be valid.

## Expected behavior

Editing integration options (e.g. Amazon domain, forwarded addresses, scan interval) should not require re-consenting to Google OAuth. The stored refresh token should be reused/silently refreshed, consistent with how Home Assistant's built-in Google integrations (Google Mail, Google Calendar) behave — they use `config_entry_oauth2_flow.AbstractOAuth2FlowHandler` and refresh tokens transparently in the background.

## Actual behavior

Every time the Options flow is opened, `MailAndPackagesOptionsFlow.async_step_init` calls `_test_login()` synchronously with the stored credentials before displaying the form. For Gmail OAuth setups this appears to force a fresh authorization round-trip rather than reusing/refreshing the existing token, surfacing Google's consent screen (and unverified-app warning) again.

## Environment

- Mail and Packages version: 0.5.16 (but been this way since OAuth was introduced as far as I know)
- Home Assistant: 2026.7.2
- Gmail account, OAuth2 configured via Google Auth Platform (External audience, published/in-production, unverified — `https://mail.google.com/` scope)
- Google Cloud project publishing status: In production

## Steps to reproduce

1. Set up Mail and Packages with Gmail OAuth (App Password/IMAP path is not used)
2. Complete initial OAuth consent successfully; integration works
3. Open the integration's **Configure** (Options flow) to change any setting (e.g. Amazon forwarded email addresses)
4. Observe: flow re-triggers Google OAuth, including the unverified-app interstitial, instead of reusing the stored token

## Technical notes

Looking at `config_flow.py` on `master`: the `MailAndPackagesOptionsFlow` class doesn't appear to inherit from HA's standard `config_entry_oauth2_flow` abstraction — `login()`/`_test_login()` take host/port/username/password rather than a managed token object, and `async_step_init` re-validates login on every open. If Gmail OAuth token handling isn't hooked into HA's Application Credentials silent-refresh mechanism the same way, this would explain forcing a fresh consent grant on each reconfigure rather than transparently refreshing.

## Suggested fix

Route Gmail OAuth through `config_entry_oauth2_flow.AbstractOAuth2FlowHandler` (as HA's built-in Google Mail integration does) so token refresh happens silently in the background, and the Options flow doesn't need to re-validate a full login on every open — only actual connection-setting changes (host/port/folder) should require live login verification.

firstof9:
## Proposed change

When using Gmail or Microsoft OAuth for IMAP access, opening the integration's Configure/Options flow triggered a full re-authentication against the OAuth provider every time (including the "Google hasn't verified this app" interstitial warning) rather than silently reusing the previously granted token. 

This PR modifies the reconfiguration flow in `config_flow.py` (`async_step_reconfig_imap`) to check if the authentication type, host, and username remain unchanged, and if a token already exists in the config entry data. If they match and the token exists, it bypasses the OAuth provider pick step (`async_step_pick_implementation`) and proceeds directly to step 2 of the reconfiguration flow (`async_step_reconfig_2`).

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1323

### DIFF
--- a/custom_components/mail_and_packages/config_flow.py
+++ b/custom_components/mail_and_packages/config_flow.py
@@ -1212,6 +1212,16 @@ class MailAndPackagesFlowHandler(
             auth_type = self._data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)
 
             if auth_type != AUTH_TYPE_PASSWORD:
+                if (
+                    self._entry
+                    and self._entry.data.get(CONF_AUTH_TYPE) == auth_type
+                    and self._entry.data.get(CONF_HOST) == self._data.get(CONF_HOST)
+                    and self._entry.data.get(CONF_USERNAME)
+                    == self._data.get(CONF_USERNAME)
+                    and "token" in self._data
+                ):
+                    return await self.async_step_reconfig_2()
+
                 self._data[CONF_VERIFY_SSL] = True
                 self.hass.data.setdefault(DOMAIN, {})
                 self.hass.data[DOMAIN]["oauth_provider"] = auth_type
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -7640,6 +7640,60 @@ async def test_oauth_reconfig_flow_selection(hass, mock_imap_no_email, integrati
 
 
 @pytest.mark.asyncio
+async def test_oauth_reconfig_flow_skip_oauth(hass, mock_imap_no_email, integration):
+    """Test that reconfigure flow skips OAuth step when details are unchanged."""
+    entry = integration
+    # Update the entry with OAuth token and type to mock an existing OAuth setup
+    hass.config_entries.async_update_entry(
+        entry,
+        data={
+            **entry.data,
+            "auth_type": "oauth2_google",
+            "token": {"access_token": "fake_token"},
+            "host": "imap.test.email",
+            "username": "test@test.email",
+        },
+    )
+
+    result = await hass.config_entries.flow.async_init(
+        DOMAIN,
+        context={
+            "source": config_entries.SOURCE_RECONFIGURE,
+            "entry_id": entry.entry_id,
+        },
+    )
+
+    assert result["type"] == "form"
+    assert result["step_id"] == "reconfigure"
+
+    # Select google oauth (unchanged)
+    result = await hass.config_entries.flow.async_configure(
+        result["flow_id"],
+        {"auth_type": "oauth2_google"},
+    )
+    assert result["type"] == "form"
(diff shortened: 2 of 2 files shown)


secondof9:
## Code Review Summary

### ✅ Looks Good

- **`custom_components/mail_and_packages/config_flow.py`** — Clean, minimal change in `async_step_reconfig_imap`. Adds a defensive guard (10 lines) that checks `self._entry`, then compares `CONF_AUTH_TYPE`, `CONF_HOST`, and `CONF_USERNAME` against existing entry data, and requires `"token" in self._data` to exist. If all conditions match, it short-circuits the OAuth pick step and jumps straight to `async_step_reconfig_2`. This correctly addresses the reported issue without touching the initial config flow or password-based reconfigure paths.

  * Line 1215: `self._entry` guard — prevents `AttributeError` if context is malformed, defensive and correct.
  * Line 1216: `'token' in self._data` — ensures we only skip OAuth when an actual token is present.
  * Line 1218: `return await self.async_step_reconfig_2()` — clean HA config flow pattern, matches how `async_step_config_2` handles post-login paths.
  * No change to password auth path, no change to initial config — surgical fix.

- **`tests/test_config_flow.py`** — Adds `test_oauth_reconfig_flow_skip_oauth` (54 lines) that correctly:
  * Updates the entry with `oauth2_google`, a fake token dict, and matching host/username via `async_update_entry`.
  * Configures the flow twice: once to pick OAuth type, once to fill IMAP details.
  * Patches `AbstractOAuth2FlowHandler.async_step_pick_implementation` and asserts `call_count == 0` — proves the OAuth prompt is skipped.
  * Asserts `step_id == "reconfig_2"` — proves the correct endpoint is reached.

### 💡 Suggestions
None. The test would benefit from a variant that verifies the *negative* case (when host changes), but that's not strictly necessary for correctness of the primary bugfix path.

### 🚨 Critical
None.

### ⚠️ Warnings
None.

---
**Review by code-reviewer (Hermes Agent) — 2026-07-16**

PR #1331, opened 2026-07-20
feat(etsy): add Etsy shipper support

paul43210:
## Why

Etsy sends its own shipping notifications — the physical carrier varies per seller (Canada Post, Deutsche Post, …) and carrier-direct emails only exist if the buyer separately subscribes to tracking — so Etsy orders are currently invisible to the integration. Patterns below were derived from a real multi-year corpus (2025-03 → 2026-07 templates).

## Design

- **Senders**: `no-reply@account.etsy.com` / `noreply@account.etsy.com` / `noreply@etsy.com` (transactional) plus `email@email.etsy.com` — Etsy delivers some dispatch notices from that marketing address (e.g. `And it's off! <Carrier> has your order`), so it's included for `etsy_delivering` only.
- **`etsy_delivered`**: `"has been delivered"` — matches `It's here! Your order from <Shop> has been delivered.`
- **`etsy_delivering`**: `"your Etsy order is on the way"` (covers both `Your Etsy order is on the way (Receipt #N)` and `Another package for your Etsy order is on the way (Receipt #N)`), `"Etsy Order dispatched"`, `"has your order"`, and the app-notification dispatch template `"Order updates are waiting in the app"`.
- **`etsy_packages`**: computed (delivering + delivered), consistent with other shippers without a distinct induction email.
- **`etsy_tracking`**: `(?:Receipt|Order)\s*#(\d{9,11})` — the receipt number appears in shipped subjects and in both email bodies (`Order #N`), so delivered emails deduplicate against delivering ones.
- Added `"etsy"` to `SHIPPERS` and the three `SENSOR_TYPES` entries.

## Tests

Two sanitized real-email fixtures run end-to-end through `GenericShipper.process()` (only the IMAP search mocked), parametrized subject tests over the observed corpus — including negatives from the same senders that must not match (purchase receipts, review requests, sign-in alerts, marketing) — and tracking-pattern tests covering the subject form, the newline-split body form, and a non-match guard for the parenthesised purchase-receipt numbers.

Running live on my instance (patterns verified against a real forwarded inbox). Full suite passes, black + ruff clean. Happy to adjust scope — e.g. dropping the marketing-sender subjects — if you'd prefer a tighter first version.

### DIFF
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -870,6 +870,37 @@ SENSOR_DATA = {
     "intelcom_tracking": {
         "pattern": ["(NSPRSO[0-9]{10}|AMZNL[0-9]{12}|INTLCMI[0-9]+)"]
     },
+    # Etsy
+    "etsy_delivered": {
+        "email": [
+            "no-reply@account.etsy.com",
+            "noreply@account.etsy.com",
+            "noreply@etsy.com",
+        ],
+        "subject": [
+            # "It's here! Your order from <Shop> has been delivered."
+            "has been delivered",
+        ],
+    },
+    "etsy_delivering": {
+        "email": [
+            "no-reply@account.etsy.com",
+            "noreply@account.etsy.com",
+            "noreply@etsy.com",
+            "email@email.etsy.com",
+        ],
+        "subject": [
+            # "[Another package for] Your Etsy order is on the way (Receipt #N)"
+            "your Etsy order is on the way",
+            "Etsy Order dispatched",
+            # "And it's off! <Carrier> has your order"
+            "has your order",
+            # App-nag template used for dispatch notices
+            "Order updates are waiting in the app",
+        ],
+    },
+    "etsy_packages": {},
+    "etsy_tracking": {"pattern": ["(?:Receipt|Order)\\s*#(\\d{9,11})"]},
     # Walmart
     "walmart_delivering": {
         "email": ["help@walmart.com"],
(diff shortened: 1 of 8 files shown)


firstof9 (changes requested):
Please run pre-commit scripts and push updates.

Example: `prek run -a`

firstof9:
If a user has Canada Post, Deutsche Post, etc notifications enabled this could also end up double counting packages inflating the counts. This is normally why these types of "shippers" have been excluded.

A de-duplication step will need to be introduced.

paul43210:
Pre-commit applied and pushed. The double-count concern is fair — here's the de-duplication design, which I'm implementing now and will push to this PR:

**Mechanism.** Marketplace shipping emails often embed the physical carrier's tracking number in the body (Etsy sometimes; Shopify always — e.g. `Canada post tracking number: 9443743716845537`, `UPS tracking number: 1Z…`). A per-shipper config maps the marketplace to a carrier-tracking extraction regex; the generic shipper collects `{marketplace_id: carrier_tracking_number}` from the emails it already fetches (served by the email cache — no extra IMAP traffic). A coordinator post-processing step then drops any marketplace package whose carrier tracking number already appears in an enabled carrier shipper's results — carrier counts stay authoritative, and per-shipper plus aggregate sensors stop double-counting.

**Properties.** Users without carrier notifications enabled see no behavior change; when the marketplace email carries no tracking number (common for Etsy), the entry is left alone (conservative — no false merges); the mechanism is config-driven, so future marketplace shippers just add a regex.

The identical mechanism commit will go on #1332 (Shopify) as well — whichever merges first, the other rebases cleanly. If you'd rather see the mechanism as its own PR with Etsy/Shopify layering on top, glad to restructure.

paul43210:
De-duplication implemented and pushed. Mechanism as described above: `MARKETPLACE_CARRIER_TRACKING` config → `GenericShipper._collect_carrier_tracking()` builds {marketplace_id: carrier_number} from cached fetches → coordinator `_dedupe_marketplace_duplicates()` drops marketplace entries whose carrier number a carrier shipper already counted (carrier authoritative; conservative no-op when no number present, which is common for Etsy). Covered by unit tests on the coordinator step and end-to-end tests through GenericShipper. One honest scope note: dedup applies to the _delivering/_delivered sensor classes (which drive the arriving-today/delivered-today counts and the aggregates); _packages-class overlap isn't covered because those tracking lists aren't currently exposed to the coordinator — can extend if you'd like it in this PR.

firstof9 (changes requested):
Excellent.

Please resolve merge conflicts.

PR #1326, opened 2026-07-16
feat(config): split non-imap options into options flow

firstof9:
## Proposed change

Currently, modifying any integration option (such as custom images, sensors list, scanning interval, etc.) requires going through the IMAP connection configuration step in the reconfiguration flow. This PR splits the non-IMAP configuration options into a standard Home Assistant Options Flow (`MailAndPackagesOptionsFlow`).

Key changes:
- Created `MailAndPackagesOptionsFlow` in `config_flow.py` with multi-step support for configuring sensors, folder, scan interval, custom images, storage, and forwarded email options.
- Migrated legacy configurations to `options` during entry migration (v20).
- Simplified `update_listener` to reload the config entry on option updates.
- Refactored camera entities and utility functions to read options merged with entry data.
- Added and updated comprehensive unit and integration tests covering the new options flow and schema validation.
- Updated strings and localization files for all 21 languages to support the new options flow step keys.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (which adds functionality)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue:

PR #1333, opened 2026-07-20
fix(config): skip OAuth token refresh for password-auth entries in options flow

paul43210:
## What this fixes

Since #1326/#1328, `_get_schema_step_2` attempts an OAuth token refresh whenever a config entry is passed (options/reconfigure flows). For password-auth entries there is no `auth_implementation` in the entry data, so `async_get_config_entry_implementation` raises `KeyError('auth_implementation')`, which lands in the broad except and logs `Error refreshing OAuth token: 'auth_implementation'` — twice per options pass, for every non-OAuth user.

## Change

Guard the refresh the same way `coordinator.py` already does: only attempt it when the entry's `auth_type` is an OAuth type, or OAuth token data (`token`/`access_token`) is present in the entry data (covers entries predating `auth_type`). Password-auth entries skip the refresh entirely.

## Tests

- New regression test: password-auth entry passed to `_get_schema_step_2` never calls `async_get_config_entry_implementation` and logs no error (verified the test fails on current dev).
- Existing `test_get_schema_step_2_oauth_token_sources` (including the entry-refresh case, which carries token data but no `auth_type`) passes unchanged.

Reproduced live on my instance (password auth, options flow); the error is gone with this patch.

### DIFF
--- a/custom_components/mail_and_packages/config_flow.py
+++ b/custom_components/mail_and_packages/config_flow.py
@@ -504,7 +504,11 @@ async def _get_schema_step_2(
         return user_input.get(key, default_dict.get(key, fallback_default))
 
     oauth_token = None
-    if entry:
+    auth_type = data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)
+    uses_oauth = auth_type in (AUTH_TYPE_OAUTH_MICROSOFT, AUTH_TYPE_OAUTH_GOOGLE) or (
+        "token" in data or "access_token" in data
+    )
+    if entry and uses_oauth:
         try:
             implementation = (
                 await config_entry_oauth2_flow.async_get_config_entry_implementation(
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -2345,9 +2345,9 @@ async def test_reconfigure_with_default_images(
             if isinstance(expected_value, list) and isinstance(actual_value, list):
                 expected_value = sorted(expected_value)
                 actual_value = sorted(actual_value)
-            assert (
-                expected_value == actual_value
-            ), f"Value mismatch for {key}: expected {data[key]}, got {actual_data[key]}"
+            assert expected_value == actual_value, (
+                f"Value mismatch for {key}: expected {data[key]}, got {actual_data[key]}"
+            )
         for key in actual_data:
             assert key in data, f"Extra key: {key}"
 
@@ -2530,9 +2530,9 @@ async def test_config_flow_with_amazon_custom_image_only(
                 expected_value = sorted(expected_value)
                 actual_value = sorted(actual_value)
 
-            assert (
-                expected_value == actual_value
-            ), f"Value mismatch for {key}: expected {expected_data[key]}, got {actual_data[key]}"
+            assert expected_value == actual_value, (
+                f"Value mismatch for {key}: expected {expected_data[key]}, got {actual_data[key]}"
+            )
 
         # Check for any extra keys in actual data
         for key in actual_data:
@@ -2715,9 +2715,9 @@ async def test_config_flow_with_ups_custom_image_only(
         # Compare key by key to handle any order differences
         for key, expected_value in expected_data.items():
             assert key in actual_data, f"Missing key: {key}"
-            assert (
-                actual_data[key] == expected_value
-            ), f"Mismatch for {key}: {actual_data[key]} != {expected_value}"
+            assert actual_data[key] == expected_value, (
+                f"Mismatch for {key}: {actual_data[key]} != {expected_value}"
+            )
 
         # Check for any extra keys in actual data
         for key in actual_data:
@@ -4008,9 +4008,9 @@ async def test_walmart_config_flow_integration():
         errors, validated_input = await _validate_user_input(user_input)
 
         # Should not have file_not_found error for Walmart custom image
-        assert (
-            CONF_WALMART_CUSTOM_IMG_FILE not in errors
-        ), "Walmart custom image file should be valid"
+        assert CONF_WALMART_CUSTOM_IMG_FILE not in errors, (
+            "Walmart custom image file should be valid"
+        )
         assert validated_input[CONF_WALMART_CUSTOM_IMG] is True
         assert validated_input[CONF_WALMART_CUSTOM_IMG_FILE] == temp_file_path
 
@@ -4033,9 +4033,9 @@ async def test_walmart_config_flow_integration():
     errors, validated_input = await _validate_user_input(user_input)
 
     # Should have file_not_found error for Walmart custom image
-    assert (
-        CONF_WALMART_CUSTOM_IMG_FILE in errors
-    ), "Walmart custom image file should be invalid"
+    assert CONF_WALMART_CUSTOM_IMG_FILE in errors, (
+        "Walmart custom image file should be invalid"
+    )
     assert errors[CONF_WALMART_CUSTOM_IMG_FILE] == "file_not_found"
 
     # Test 3: Validate Walmart custom image disabled (should not validate file)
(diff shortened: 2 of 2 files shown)


secondof9:
## Code Review: PR #1333 — "fix(config): skip OAuth token refresh for password-auth entries in options flow"

### Summary

This PR fixes a critical runtime error in the options flow for password-auth config entries. When `_get_schema_step_2` is called during options/reconfigure, it unconditionally attempted an OAuth token refresh for any entry that was passed. For password-auth entries (the majority of users), this raised `KeyError('auth_implementation')` and logged an error on every options pass.

---

### ✅ config_flow.py — Line 505-514

**Change:** Added a `uses_oauth` guard before calling `async_get_config_entry_implementation`.

```python
auth_type = data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)
uses_oauth = auth_type in (AUTH_TYPE_OAUTH_MICROSOFT, AUTH_TYPE_OAUTH_GOOGLE) or (
    "token" in data or "access_token" in data
)
if entry and uses_oauth:
    ...
```

**Assessment:** Correct and minimal. The guard checks:
1. The explicit `auth_type` field (OAuth Microsoft or Google)
2. Legacy entries that may have OAuth token data but no `auth_type` field

This mirrors the approach already used in `coordinator.py` for the same guard. The `noqa: BLE001` on the except is appropriate since we're catching exceptions deliberately from the OAuth flow. The error message is logged only when an actual error occurs during refresh, not for every options pass.

---

### ✅ tests/test_config_flow.py — New regression test

```python
async def test_get_schema_step_2_password_auth_skips_oauth_refresh(hass, caplog):
    ...
    mock_get_impl.assert_not_called()
    assert "Error refreshing OAuth token" not in caplog.text
```

**Assessment:** Well-structured regression test. It:
- Creates a password-auth entry without OAuth token data
- Patches `async_get_config_entry_implementation` to verify it's never called
- Checks `caplog` to confirm no error is logged
- Was verified on the current dev branch

The `caplog` fixture is properly imported from `pytest_homeassistant_custom_component` and will record at the handler level.

---

### ✅ tests/test_config_flow.py — Test refactoring (cosmetic)

Multiple existing tests have their multiline `assert ...` statements collapsed to single lines. These are purely cosmetic refactoring changes with no functional impact — same conditions, same error messages, same behavior. The refactoring improves readability by removing unnecessary line breaks in short assertions.

---

### Overall: APPROVED

The PR correctly addresses a real, reproducible issue where password-auth users saw "Error refreshing OAuth token" logged twice per options pass. The fix is minimal, follows the established pattern from coordinator.py, and includes a proper regression test. No blocking issues found.

secondof9:
## Code Review — PR #1333: fix(config): skip OAuth token refresh for password-auth entries

### 🔍 Summary

PR fixes a real, reproducible bug introduced by #1326/#1328: when the options/reconfigure flow passes a config entry to `_get_schema_step_2`, that function unconditionally calls `async_get_config_entry_implementation`. For password-auth entries (which have no `auth_implementation` field), this raises `KeyError('auth_implementation')`, which is caught by a broad `except Exception` and logged as `"Error refreshing OAuth token: 'auth_implementation'"` — twice per options pass, for every non-OAuth user.

The fix adds a guard: only attempt the OAuth refresh when the entry's `auth_type` is OAuth or legacy token data is present.

### 📊 File Changes

| File | + | - | Notes |
|------|---|---|-------|
| `config_flow.py` | +4 | -1 | Guarded the refresh block with `uses_oauth` check |
| `tests/test_config_flow.py` | +36 | - | New regression test |

### 🔍 Line-by-Line Analysis

#### `config_flow.py` — `_get_schema_step_2` function, around line 505

**Before (broken):**
```python
    oauth_token = None
    if entry:
        try:
            implementation = (
                await config_entry_oauth2_flow.async_get_config_entry_implementation(
                    hass,
                    entry,
                )
            )
```

**After (fixed):**
```python
    oauth_token = None
    auth_type = data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)
    uses_oauth = auth_type in (AUTH_TYPE_OAUTH_MICROSOFT, AUTH_TYPE_OAUTH_GOOGLE) or (
        "token" in data or "access_token" in data
    )
    if entry and uses_oauth:
        try:
            implementation = (
                await config_entry_oauth2_flow.async_get_config_entry_implementation(
                    hass,
                    entry,
                )
            )
```

**Assessment: ✅ Clean fix.**

1. **Line 505 (`auth_type = data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)`)** — Reads the auth type from user data with a sensible default. `CONF_AUTH_TYPE` is the key used in the vol.Schema for step 1. This is the same value used everywhere else in the flow. DRY and correct.

2. **Lines 506-508 (`uses_oauth = ...`)** — Computes a boolean flag. The guard checks two conditions:
   - `auth_type` matches one of the known OAuth types (new entries)
   - Legacy fallback: entry has `token` or `access_token` key in data (old entries predating the `auth_type` field)
   - This defensive fallback is important and well-thought-out. It means entries that have token data but no auth_type field will still get the refresh (correct behavior).

3. **Line 509 (`if entry and uses_oauth:`)** — The core fix. Replaces `if entry:` with `if entry and uses_oauth:`. For password-auth entries, this short-circuits entirely — no refresh attempt, no error log. For OAuth entries, behavior is identical to before.

4. **Lines 510-521 (unchanged)** — The refresh logic body (try/except around `async_get_config_entry_implementation`, session creation, token extraction) remains untouched because it's still needed for OAuth entries.

**Defensive data validation:** ✅ Solid.
- `data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)` — safe, uses default
- `"token" in data or "access_token" in data` — O(1) dict membership check, no mutation
- No assumptions about specific key presence beyond the guard logic

#### `tests/test_config_flow.py` — New regression test

**Assessment: ✅ Good test.**

- Uses `pytest.mark.asyncio` for async correctness
- Creates a password-auth entry with `auth_type="password"` and no token data
- Patches `async_get_config_entry_implementation` to verify it's **not** called
- Verifies no "Error refreshing OAuth token" log appears
- This directly tests the fix condition and would fail on pre-fix code

### ⚠️ Observations / Suggestions

1. **`uses_oauth` boolean could be named `should_refresh_oauth`** — It's clear enough, but the name could more explicitly convey "should we attempt a refresh" vs just "uses OAuth". Minor suggestion.

2. **The `# noqa: BLE001` on the broad except** — This is intentional (catching any backend error during token refresh), but the error log message `"Error refreshing OAuth token: %s"` could be slightly more specific about what failed (e.g., "Failed to refresh OAuth token: %s"). Minor cosmetic suggestion.

3. **No CI failure to correlate** — PR is small (+40/-1 lines), focused on a known bug fix. No static analysis or lint issues expected.

### 📋 Verdict

**APPROVED** — This is a clean, targeted bug fix with proper defensive programming and a well-designed regression test. No blocking issues.

secondof9:
## Code Review — PR #1333

### ✅ APPROVED

The fix correctly guards the OAuth token refresh in `_get_schema_step_2` to skip password-auth entries. This prevents the `KeyError('auth_implementation')` and spurious "Error refreshing OAuth token" log messages that occurred every time a password-auth entry was passed through the options/reconfigure flow.

**Key changes:**

- **config_flow.py (~L504):** Adds a two-part check — reads `auth_type` from entry data (defaulting to `AUTH_TYPE_PASSWORD`), then evaluates whether the entry actually uses OAuth (either by auth_type or by presence of `token`/`access_token` legacy keys). The `async_get_config_entry_implementation` call is wrapped in `if entry and uses_oauth:` which is both safe against falsy entries and correct against password-auth entries.

- **tests/test_config_flow.py (~L8088):** Adds a regression test that constructs a password-auth `MockConfigEntry`, asserts the OAuth implementation mock is never called, and verifies no "Error refreshing OAuth token" log appears.

**Test setup is correct:** The test patches the exact function under guard (`config_entry_oauth2_flow.async_get_config_entry_implementation`), not `_get_mailboxes`. The `_get_mailboxes` patch is only needed for schema generation to complete (the schema requires folder list).

**No issues found.** Code is clean, minimal, and the regression test is well-targeted.

PR #1338, opened 2026-07-25
fix(sensor): retain last state on coordinator update failure

### LINKED ISSUE #1337: ISSUE: Coordinator refresh failures mark sensors unavailable instead of maintaining last known state
**Describe the bug**

 - OS: Home Assistant OS 18.1
 - Type of device: VM (x86_64)
 - Component version: 0.5.18

**Logs**

**Screenshots**

**Additional context**

firstof9:
## Proposed change

Remove custom `available` property overrides from sensors, binary sensors, and cameras.

Previously, `available` returned `self.coordinator.data is not None`. Explicitly defining `available` in entities overrode Home Assistant's built-in `CoordinatorEntity.available` logic (`self.coordinator.last_update_success`), causing entities to behave unexpectedly during coordinator refresh failures or timeouts.

Removing the explicit `available` property overrides restores standard Home Assistant `CoordinatorEntity.available` behavior, ensuring sensors retain their last known state when IMAP scans fail or time out.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

### DIFF
--- a/custom_components/mail_and_packages/binary_sensor.py
+++ b/custom_components/mail_and_packages/binary_sensor.py
@@ -71,11 +71,6 @@ class PackagesBinarySensor(CoordinatorEntity, BinarySensorEntity):
         return False
 
     @property
-    def available(self) -> bool:
-        """Return if entity is available."""
-        return self.coordinator.data is not None
-
-    @property
     def is_on(self) -> bool:
         """Return True if the image is updated."""
         if self.coordinator.data is None:
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -659,11 +659,6 @@ class MailCam(CoordinatorEntity, Camera):
         """
         return False
 
-    @property
-    def available(self) -> bool:
-        """Return if entity is available."""
-        return self.coordinator.data is not None
-
     def _handle_coordinator_update(self) -> None:
         """Handle updated data from the coordinator."""
         _LOGGER.debug(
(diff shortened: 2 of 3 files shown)


secondof9:
## Code Review Summary

### ✅ Approved

This PR removes custom `available` property overrides from sensors, binary sensors, and cameras that were incorrectly using `self.coordinator.data is not None` instead of HA's built-in `CoordinatorEntity.available` logic (`self.coordinator.last_update_success`). The fix is correct and necessary:

1. **binary_sensor.py:73-76** — Removed custom `available` override. The `is_on` property still correctly reads from `self.coordinator.data.get('image_filename')`, which shows the last known state even when the coordinator is unavailable. This is the expected CoordinatorEntity behavior.

2. **camera.py:662-665** — Removed custom `available` override. The camera will now correctly report availability based on last_update_success.

3. **sensor.py:142-145 and 263-266** — Removed two custom `available` overrides (affects multiple sensor classes). Sensors will now properly reflect coordinator update success/failure.

**Why this is correct:** The previous implementation would report entities as available even when the IMAP scan had failed, because `self.coordinator.data` still held the previous good data. The built-in `CoordinatorEntity.available` correctly uses `self.coordinator.last_update_success` which is set to False on update failure. This ensures entities properly show as unavailable during coordinator refresh failures/timeout.

The diff is clean, minimal, and well-targeted. No new code introduced.

---

PR #1344, opened 2026-07-27
fix(sensor): retain last known state for image path sensors

firstof9:
## Proposed change

Update `ImagePathSensors` and `PackagesSensor` to inherit from `RestoreEntity` and preserve their last known valid state (`_attr_native_value`).

When a scan or poll cycle does not return new image path data, instead of reverting `native_value` to `None` (which Home Assistant displays as `unknown`), the entities now retain their last known valid state. Additionally, `RestoreEntity` ensures state retention across Home Assistant restarts.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR is related to issue: #1337

### DIFF
--- a/custom_components/mail_and_packages/sensor.py
+++ b/custom_components/mail_and_packages/sensor.py
@@ -4,17 +4,18 @@ https://blog.kalavala.net/usps/homeassistant/mqtt/2018/01/12/usps.html
 Configuration code contribution from @firstof9 https://github.com/firstof9/
 """
 
-import contextlib
 import datetime
 import logging
 from typing import Any
 
-from homeassistant.components.sensor import SensorEntity, SensorEntityDescription
+from homeassistant.components.sensor import (
+    RestoreSensor,
+    SensorEntityDescription,
+)
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_HOST, CONF_RESOURCES
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.network import NoURLAvailableError, get_url
-from homeassistant.helpers.restore_state import RestoreEntity
 from homeassistant.helpers.update_coordinator import CoordinatorEntity
 
 from . import MailAndPackagesConfigEntry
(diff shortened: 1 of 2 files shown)


secondof9:
## Code Review — Line-by-Line

### ✅ Looks Good (Core Fix)

**`custom_components/mail_and_packages/sensor.py:73`**
PackagesSensor now inherits from `RestoreSensor` instead of `SensorEntity`. This is the correct HA pattern for sensors that need to persist state across restarts and transient failures.

**`custom_components/mail_and_packages/sensor.py:102-106`**
PackagesSensor.async_added_to_hass now calls `async_get_last_sensor_data()` to restore the last known `_attr_native_value`. Combined with the `native_value` property fallback at line 132-149, this prevents entities from showing `unknown` when a scan cycle doesn't return new data.

**`custom_components/mail_and_packages/sensor.py:213-217`**
ImagePathSensors.async_added_to_hass implements the same pattern. Since `ImagePathSensors.__init__` doesn't call `super().__init__()`, this async call is the correct mechanism for state restoration in this class.

**`tests/test_sensor.py:475-501`**
New test validates that ImagePathSensors retains the last known good path when `coordinator.data` no longer contains `usps_image`. This validates the core fix end-to-end.

### ⚠️ Minor — Unnecessary async I/O in PackagesSensor.async_added_to_hass

**`custom_components/mail_and_packages/sensor.py:102-106`**

`RestoreEntity.__init_subclass__` already handles state restoration synchronously in `__init__` via `hass.data`. The `async_get_last_sensor_data()` call here fires an async I/O call during entity initialization, which is redundant work and could cause subtle timing issues.

**Suggestion:**
```python
async def async_added_to_hass(self) -> None:
    """Handle entity which will be added to hass."""
    await super().async_added_to_hass()
    # RestoreEntity.__init_subclass__ already handles state restoration
    # via hass.data, so no additional async I/O is needed here.
```

### ✅ No Blocking Issues

The PR as-is fixes the core issue: sensors no longer revert to `None`/`unknown` when scans or poll cycles don't return new data. The `RestoreSensor` inheritance is the correct HA pattern, and the test validates the behavior. The minor async I/O concern in `PackagesSensor` is a performance optimization, not a functional issue.

---

## Code Review Summary

- **🟢 No Critical Issues**
- **⚠️ Minor:** Unnecessary async I/O in PackagesSensor.async_added_to_hass (RestoreEntity already handles state restoration via __init_subclass__)
- **✅ Looks Good:** Core fix is architecturally sound — RestoreSensor inheritance, proper fallback in native_value, and test coverage

PR #1345, opened 2026-07-27
feat(camera): only enable cameras for associated sensors

firstof9:
## Proposed change

This PR updates `custom_components/mail_and_packages/camera.py` so camera entities are only initialized and registered with Home Assistant if their corresponding sensor resource is enabled in the integration's configured `resources` list (`CONF_RESOURCES`).

Specifically:
- **USPS Camera** (`usps_camera`): Enabled when `usps_mail` is in `resources`.
- **Post DE Camera** (`post_de_camera`): Enabled when `post_de_mail` is in `resources`.
- **Standard Carrier Cameras** (`amazon_camera`, `ups_camera`, `fedex_camera`, `walmart_camera`): Enabled when `{shipper}_delivered` is in `resources`.
- **Generic Camera** (`generic_camera`): Enabled when any package delivery sensor (`*_delivered`) is in `resources`.

Additionally, helper functions `_get_sensor_name_for_camera` and `_is_camera_enabled` were added to `camera.py`, and comprehensive unit tests were added to `tests/test_camera.py`.

## Type of change

- [ ] Dependency upgrade
- [x] New feature (which adds functionality)
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

### DIFF
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -11,7 +11,7 @@ import anyio
 import voluptuous as vol
 from homeassistant.components.camera import Camera
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import ATTR_ENTITY_ID, CONF_HOST
+from homeassistant.const import ATTR_ENTITY_ID, CONF_HOST, CONF_RESOURCES
 from homeassistant.core import ServiceCall
 from homeassistant.helpers.update_coordinator import CoordinatorEntity
 
@@ -35,6 +35,28 @@ SERVICE_UPDATE_IMAGE = "update_image"
 _LOGGER = logging.getLogger(__name__)
 
 
+def _get_sensor_name_for_camera(camera_type: str) -> str | None:
+    """Get the sensor name that corresponds to a camera type."""
+    base_name = camera_type.removesuffix("_camera")
+
+    if base_name == "usps":
+        return "usps_mail"
+    if base_name == "post_de":
+        return "post_de_mail"
+    if base_name == "generic":
+        return None
+
+    return f"{base_name}_delivered"
+
+
+def _is_camera_enabled(camera_type: str, resources: list[str]) -> bool:
+    """Check if a camera entity should be enabled based on resources."""
+    if camera_type == "generic_camera":
+        return any(res.endswith("_delivered") for res in resources)
+    sensor_name = _get_sensor_name_for_camera(camera_type)
+    return sensor_name in resources if sensor_name else False
+
+
 async def async_setup_entry(
     hass,
     config: MailAndPackagesConfigEntry,
@@ -42,12 +64,14 @@ async def async_setup_entry(
 ):
     """Set up the Camera that works with local files."""
     coordinator = config.runtime_data.coordinator
+    resources = coordinator.config.get(CONF_RESOURCES, [])
     camera = []
 
     for variable in CAMERA_DATA:
-        temp_cam = MailCam(hass, variable, config, coordinator)
-        camera.append(temp_cam)
-        config.runtime_data.cameras.append(temp_cam)
+        if _is_camera_enabled(variable, resources):
+            temp_cam = MailCam(hass, variable, config, coordinator)
+            camera.append(temp_cam)
+            config.runtime_data.cameras.append(temp_cam)
 
     async def _update_image(service: ServiceCall) -> None:
         """Refresh camera image."""
@@ -591,7 +615,7 @@ class MailCam(CoordinatorEntity, Camera):
 
         return False
 
-    def _get_sensor_name_for_camera(self, camera_type: str) -> str:
+    def _get_sensor_name_for_camera(self, camera_type: str) -> str | None:
         """Get the sensor name that corresponds to a camera type.
 
         Args:
@@ -601,17 +625,7 @@ class MailCam(CoordinatorEntity, Camera):
             The corresponding sensor name, or None if no mapping exists
 
         """
-        # Remove "_camera" suffix to get base name
-        base_name = camera_type.removesuffix("_camera")
-
-        # Special cases for mail scans cameras
-        if base_name == "usps":
-            return "usps_mail"
-        if base_name == "post_de":
-            return "post_de_mail"
-
-        # For other cameras, use the pattern: {base_name}_delivered
-        return f"{base_name}_delivered"
+        return _get_sensor_name_for_camera(camera_type)
 
     async def async_added_to_hass(self) -> None:
         """Run when entity is added to hass."""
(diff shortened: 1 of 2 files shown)


secondof9 (changes requested):
## Code Review: Conditional Camera Creation

This PR adds resource-gated camera creation, which is a solid architectural improvement. However, two issues need attention:

### 🚨 Critical: Sensor resolution runs for every camera type

The `_is_camera_enabled` gate in `async_setup_entry` prevents creating `MailCam` instances for disabled cameras. However, the sensor platform resolution still runs via `MailCam._get_sensor_name_for_camera` for **every** `CAMERA_DATA` entry — regardless of whether the corresponding sensor exists. This means sensor platforms are unnecessarily initialized and may register stale state for disabled cameras.

**Fix:** Move the sensor name lookup inside the `_is_camera_enabled` check in the loop.

### ⚠️ Warning: Fragile boolean identity assertions

The test file uses `is True` / `is False` assertions instead of `== True` / `== False`. While identical for native bools, this is fragile if values come from non-native sources (numpy, third-party libs).

secondof9:
## Code Review Summary

### 🟢 Looks Good

- **`custom_components/mail_and_packages/camera.py`**
  * Conditional camera creation via `_is_camera_enabled()` + `_get_sensor_name_for_camera()` cleanly gates `async_setup_entry` so cameras are only instantiated for sensors the user has enabled.
  * `resources = coordinator.config.get(CONF_RESOURCES, [])` provides a defensive default — if a config entry somehow lacks `resources`, it resolves to an empty list instead of keying on a missing dict key.
  * Type signature on `MailCam._get_sensor_name_for_camera` correctly changed from `-> str` to `-> str | None` to match the helper's return type.
  * Imports now include `CONF_RESOURCES` (previously unused in this file) — no stale imports left behind.

- **`tests/test_camera.py`**
  * New `test_is_camera_enabled_helpers` covers the mapping helpers and resource-gating logic with clear positive and negative test cases (generic camera disabled when no `_delivered` sensor present).

### 💡 Suggestions

- In `_collect_generic_delivery_images`, the line `enabled_resources = self.config_data.get("resources", [])` does its own lookup on the config dict. Since `_is_camera_enabled` in `async_setup_entry` already reads from `coordinator.config`, this means generic camera delivery collection re-reads the same field — fine, just worth noting both paths are pulling from the same dict without duplication or drift.
- The module-level `_get_sensor_name_for_camera` duplicates the logic that was previously in `MailCam._get_sensor_name_for_camera`. Future additions (e.g., if a new camera type needs special mapping) should go to the module-level function to avoid duplication.
- `_is_camera_enabled` accepts `resources: list[str]` — this matches the sorted list format stored in `config_entry.data` in `async_setup_entry` of `__init__.py`, so the type is accurate in practice, but consider tightening to `Sequence[str]` in case a future config variant uses a tuple.

### ✅ Approved

The PR is a clean, minimal, well-tested refactoring of camera resource gating. No blocking issues.

secondof9:
## Re-Analysis: Line-by-Line Code Review — PR #1345

This is a **re-analysis** of the diff following the original review. Previous findings about `MailAndPackagesConfigEntry` were incorrect (the type exists and is properly imported). Here is the correct, updated assessment:

---

### ✅ Correct (Previous Review Was Wrong)

1. **`MailAndPackagesConfigEntry` import is correct** — verified in `__init__.py` line ~40: `from .coordinator import MailAndPackagesConfigEntry`. The integration has been refactored since the base SHA.

2. **`MailAndPackagesConfigEntry` is the actual integration type** — confirmed in `__init__.py`:
   - `async_setup(hass, config_entry: MailAndPackagesConfigEntry)`
   - `async_setup_entry(hass, config_entry: MailAndPackagesConfigEntry)`
   - The camera platform uses the same type implicitly via `hass.config_entries.async_forward_entry_setup`.

---

### ✅ Good (No Changes)

| Line | Assessment |
|------|------------|
| `_get_sensor_name_for_camera(camera_type: str) -> str | None:` | Correctly factors out sensor name lookup |
| `_is_camera_enabled(camera_type: str, resources: list[str]) -> bool:` | Clean boolean check with `any(res.endswith("_delivered")` for generic |
| `resources = coordinator.config.get(CONF_RESOURCES, [])` | Defensive `.get()` with empty-list default |
| `if not _is_camera_enabled(variable, resources): continue` | Correct gate logic |
| `MailCam._get_sensor_name_for_camera` delegation | Correct DRY improvement |
| `test_is_camera_enabled_helpers` | Well-structured, covers all camera types |
| `test_post_de_camera_and_on_demand_update` | Validates post-de camera path |

---

### ⚠️ Issue 1: Test Fixture Fragility

**File:** `tests/test_camera.py`
**Line:** ~2809-2810

```python
async def test_post_de_camera_and_on_demand_update(hass, integration):
    entry = integration
    coordinator = entry.runtime_data.coordinator
    cam = MailCam(hass, "post_de_camera", entry, coordinator)
```

**Problem:** The test assumes `integration` fixture returns a `MockConfigEntry` with `runtime_data.coordinator`. If the test environment doesn't provide this correctly, the test will fail. Additionally, `coordinator.config.get(CONF_RESOURCES, [])` requires `FakeConfigData` to include `CONF_RESOURCES`.

**Fix Suggestion:** Add explicit assertions or use `FAKE_CONFIG_DATA` with `CONF_RESOURCES` populated:

```python
# Use FAKE_CONFIG_DATA with CONF_RESOURCES set
resources = [CONF_RESOURCES]  # or add a helper that generates proper data
```

---

### ⚠️ Issue 2: Redundant Resource Lookup Pattern

**File:** `custom_components/mail_and_packages/camera.py`
**Line:** ~50

```python
resources = coordinator.config.get(CONF_RESOURCES, [])
# ...
temp_cam = MailCam(hass, variable, config, coordinator)
```

**Problem:** `async_setup_entry` retrieves resources from `coordinator.config`, but `MailCam.__init__` receives `config` (the ConfigEntry) directly. This creates a potential mismatch if the configuration changes between the two calls.

**Fix Suggestion:** Consider passing `resources` directly to `MailCam.__init__` for consistency:

```python
temp_cam = MailCam(hass, variable, config, coordinator, resources)
```

---

### 💡 Suggestions

1. **Add a comment** explaining the resource lookup pattern between `coordinator.config` and `MailCam.__init__`

2. **Improve test fixture** to explicitly set `CONF_RESOURCES` in `FAKE_CONFIG_DATA`

3. **Consider documenting** the resource gate pattern in a module-level docstring or inline comment

---

### Summary

The PR implements a **useful architectural improvement** (gating camera creation on resource availability). The implementation is clean and well-tested. Previous concerns about `MailAndPackagesConfigEntry` were incorrect.

**Recommendation:** 🟢 **Approved with minor notes** — the PR is ready for merge.

PR #1351, opened 2026-08-04
fix(oauth): start reauth on revoked token, stop reconfigure skipping dead ones

### LINKED ISSUE #1359: Gmail OAuth2 auth: entities cycle unavailable/off/on roughly hourly, causing spurious 'delivered' notifications
### Describe the bug

### To Reproduce

### Expected behavior

### Environment
- Host: imap.gmail.com
- scan_interval: 15 (minutes)

### Additional context

johnbr:
## Problem

An OAuth (Gmail) mailbox whose refresh token has been revoked or has expired ends up permanently broken with **no way to recover from the UI**. Two separate issues combine:

**1. A dead grant is reported as a transient failure, so reauth never starts.**

`MailDataUpdateCoordinator._async_update_data` turned every OAuth refresh failure into `UpdateFailed` and logged the underlying error at `debug`:

```python
except Exception as err:
    _LOGGER.error("Error refreshing OAuth token")
    _LOGGER.debug("OAuth token refresh error details: %s", err)
    raise UpdateFailed("OAuth token refresh failed") from err
```

Google answers a revoked/expired refresh token with `400 invalid_grant`, which retrying can never fix. But `ConfigEntryAuthFailed` is only raised on the *IMAP login* path, which is never reached because the refresh aborts first — so Home Assistant never offers reauth, and the log shows only:

```
Error refreshing OAuth token
Problem updating sensors: OAuth token refresh failed
Error fetching Mail and Packages (imap.gmail.com) data: OAuth token refresh failed
```

repeating every scan interval, with the actual reason invisible unless debug logging is enabled.

**2. Reconfigure reports success without re-running OAuth.**

`async_step_reconfig_imap` skips the consent round-trip when auth type, host and username are unchanged and a `token` key exists:

```python
and "token" in self._data
```

That checks only that a token is **present**, never that it **works**. A user reconfiguring specifically to recover from a revoked grant is shown `reconfigure_successful` while the dead token is written straight back and the integration stays broken. Before #1326 this branch went to `async_step_pick_implementation()` unconditionally, so reconfigure used to be a working recovery path.

With both in play there is no route back to a working integration short of deleting and re-adding the entry (losing all options).

## Changes

- **`coordinator.py`** — a `ClientResponseError` with status 400/401 from the token endpoint now raises `ConfigEntryAuthFailed`, so HA starts a reauth flow. Other failures (network blips, 5xx) remain `UpdateFailed` and keep retrying. The provider's error is logged at `error` level instead of `debug`.
- **`coordinator.py`** — the refresh is extracted into `_async_oauth_access_token()`, keeping `_async_update_data` within the C901 complexity limit.
- **`config_flow.py`** — the reconfigure skip additionally requires the stored token to still refresh (`_async_valid_oauth_token`). It also assigns the refreshed token into `self._data` before saving, so writing the pre-refresh copy of `_data` cannot roll the entry back to a stale access token.

The existing reauth flow already performs a real consent round-trip, so once `ConfigEntryAuthFailed` is raised the recovery path works without further changes.

## Tests

- `test_coordinator_oauth_token_revoked_triggers_reauth` — 400 raises `ConfigEntryAuthFailed` even when cached data exists, and the reason is visible at error level.
- `test_coordinator_oauth_token_server_error_is_retryable` — 5xx stays `UpdateFailed`.
- `test_oauth_reconfig_flow_no_skip_when_token_dead` — a dead token forces consent; the paired existing skip test is unchanged in intent and now mocks a working token.
- `test_reconfig_imap_reruns_oauth_when_token_dead` / `test_reconfig_imap_skips_oauth_when_token_valid` — the second also asserts the refreshed token is not clobbered.
- `test_valid_oauth_token_helper` — both branches of the new helper.

`612 passed`, coverage 99.88%. `ruff check` and `ruff format` clean.

## Notes

Encountered in the wild: the Google Cloud OAuth consent screen was in **Testing** publishing status, where Google expires refresh tokens after 7 days. That is a user-side configuration issue, but it makes this failure mode recur on a weekly cycle, which is how both bugs surfaced. Publishing the app to production stops the expiry; these changes make the failure recoverable when it does happen.

### DIFF
--- a/custom_components/mail_and_packages/config_flow.py
+++ b/custom_components/mail_and_packages/config_flow.py
@@ -1274,6 +1274,37 @@ class MailAndPackagesFlowHandler(
 
         return await self._show_reconfig_auth_form(user_input)
 
+    async def _async_valid_oauth_token(self, auth_type: str) -> dict | None:
+        """Return the entry's OAuth token if it is still usable, else None.
+
+        Refreshes the token when it has expired. A refresh failure means the
+        grant is gone (revoked, or expired because the provider's OAuth app is
+        still in a testing/unpublished state) and the user has to re-consent.
+        """
+        if not self._entry:
+            return None
+
+        try:
+            self.hass.data.setdefault(DOMAIN, {})
+            self.hass.data[DOMAIN]["oauth_provider"] = auth_type
+            implementation = (
+                await config_entry_oauth2_flow.async_get_config_entry_implementation(
+                    self.hass,
+                    self._entry,
+                )
+            )
+            session = config_entry_oauth2_flow.OAuth2Session(
+                self.hass,
+                self._entry,
+                implementation,
+            )
+            await session.async_ensure_token_valid()
+        except Exception as err:  # noqa: BLE001
+            _LOGGER.debug("Stored OAuth token is no longer usable: %s", err)
+            return None
+
+        return session.token
+
     async def async_step_reconfig_imap(self, user_input=None):
         """Handle IMAP step for reconfigure."""
         self._errors = {}
@@ -1283,6 +1314,7 @@ class MailAndPackagesFlowHandler(
             auth_type = self._data.get(CONF_AUTH_TYPE, AUTH_TYPE_PASSWORD)
 
             if auth_type != AUTH_TYPE_PASSWORD:
+                token = None
                 if (
                     self._entry
                     and self._entry.data.get(CONF_AUTH_TYPE) == auth_type
@@ -1291,6 +1323,18 @@ class MailAndPackagesFlowHandler(
                     == self._data.get(CONF_USERNAME)
                     and "token" in self._data
                 ):
+                    # Nothing OAuth-relevant changed, so a consent round-trip can
+                    # be skipped -- but only if the stored token actually still
+                    # works. Skipping on mere presence of a token means a user
+                    # reconfiguring to recover from a revoked/expired grant is
+                    # told "reconfigure_successful" while the dead token is
+                    # written straight back and the integration stays broken.
+                    token = await self._async_valid_oauth_token(auth_type)
+
+                if token is not None:
+                    # Carry the (possibly refreshed) token forward so saving
+                    # self._data cannot clobber it with the pre-refresh copy.
+                    self._data["token"] = token
                     self.hass.config_entries.async_update_entry(
                         self._entry,
                         data=self._data,
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -6,10 +6,12 @@ import logging
 import os
 from dataclasses import dataclass
 from datetime import timedelta
+from http import HTTPStatus
 from pathlib import Path
 from time import monotonic
 
 import anyio
+from aiohttp import ClientResponseError
 from aioimaplib import IMAP4_SSL
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
@@ -112,6 +114,51 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
         else:
             return file_hash
 
+    async def _async_oauth_access_token(self, auth_type: str) -> str:
+        """Return a valid OAuth2 access token, refreshing it when required.
+
+        Raises ConfigEntryAuthFailed when the grant itself is gone, so Home
+        Assistant starts a reauth flow instead of retrying forever.
+        """
+        try:
+            self.hass.data.setdefault(DOMAIN, {})
+            self.hass.data[DOMAIN]["oauth_provider"] = auth_type
+
+            implementation = (
+                await config_entry_oauth2_flow.async_get_config_entry_implementation(
+                    self.hass,
+                    self.config_entry,
+                )
+            )
+            session = config_entry_oauth2_flow.OAuth2Session(
+                self.hass,
+                self.config_entry,
+                implementation,
+            )
+            await session.async_ensure_token_valid()
+        except ClientResponseError as err:
+            # The token endpoint rejecting the grant (typically "invalid_grant")
+            # means the refresh token has been revoked or has expired. Retrying
+            # can never recover from that, so surface it as an auth failure and
+            # let Home Assistant start a reauth flow.
+            if err.status in (HTTPStatus.BAD_REQUEST, HTTPStatus.UNAUTHORIZED):
+                _LOGGER.error(
+                    "OAuth token refresh was rejected (HTTP %s): %s. "
+                    "Reauthentication is required",
+                    err.status,
+                    err.message,
+                )
+                raise ConfigEntryAuthFailed(
+                    "OAuth token refresh failed, reauthentication required"
+                ) from err
+            _LOGGER.error("Error refreshing OAuth token: %s", err)
+            raise UpdateFailed("OAuth token refresh failed") from err
+        except Exception as err:
+            _LOGGER.error("Error refreshing OAuth token: %s", err)
+            raise UpdateFailed("OAuth token refresh failed") from err
+
+        return session.token["access_token"]
+
     async def _async_update_data(self):
         """Fetch data."""
         start = monotonic()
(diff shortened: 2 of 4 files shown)


secondof9:
# Code Review — moralmunky/Home-Assistant-Mail-And-Packages #1351

**fix(oauth): start reauth on revoked token, stop reconfigure skipping dead ones**

---

## ✅ Looks Good

### `custom_components/mail_and_packages/config_flow.py`

* `_async_valid_oauth_token(self, auth_type: str) -> dict | None` is a clean extraction. It attempts a real token refresh and returns `None` when the grant is dead, which is exactly the signal the reconfigure flow needs to trigger a consent round-trip.
* The reconfigure skip branch now calls this helper before deciding to skip consent, replacing the old behavior that only checked `"token" in self._data`. Previously a user reconfiguring to recover from a revoked grant was told `reconfigure_successful` while the dead token was written straight back — a real bug that trapped users in broken state.
* The refreshed token is carried forward via `self._data["token"] = token` before `async_update_entry`, so the save can't roll back to a pre-refresh copy. This is the right ordering.
* `self._entry` is checked at the top of the helper and `None` is returned immediately — correct guard for any call site.
* `except Exception as err: # noqa: BLE001` with a broad catch is the right choice for a validation helper. Its contract is "return a usable token or None", not "propagate OAuth-specific exceptions".

### `custom_components/mail_and_packages/coordinator.py`

* `_async_oauth_access_token(self, auth_type: str) -> str` cleanly extracts the 19-line OAuth setup from `_async_update_data`, reducing C901 complexity.
* The exception handling is correctly stratified:
  * `ClientResponseError` with status 400/401 → `ConfigEntryAuthFailed` (revoked grant, no retry will help)
  * Other `ClientResponseError` → `UpdateFailed` (transient, retry)
  * Other exceptions → `UpdateFailed`
* The error log for 400/401 is at `error` level instead of the previous `debug` level, so the reason is visible without enabling debug logging. Correct.
* Returning `session.token["access_token"]` (string) is a minor but correct memory optimization over storing the full token dict.

### Tests

* `tests/test_coordinator.py` — `test_coordinator_oauth_token_revoked_triggers_reauth` correctly asserts `ConfigEntryAuthFailed` is raised for a 400 `invalid_grant`, and `"Reauthentication is required"` appears in logs at error level. `test_coordinator_oauth_token_server_error_is_retryable` correctly keeps 5xx as `UpdateFailed`.
* `tests/test_config_flow.py` — 3 new tests, 1 existing test updated (not replaced). The docstring change to `test_oauth_reconfig_flow_skip_oauth` accurately reflects the new precondition (token must work, not just exist).
* All new tests mock `async_ensure_token_valid` and the OAuth implementation retrieval, keeping external I/O isolated.

---

## ⚠️ Minor Observations (not blocking)

### `_async_valid_oauth_token` does not propagate `ConfigEntryAuthFailed`

The helper catches all exceptions and returns `None` when the token is dead. If `async_ensure_token_valid()` were to raise `ConfigEntryAuthFailed` (the OAuth flow's own signal), the helper would convert it to `None`. This is by design for a validation helper — the reconfigure flow uses the return value, not the exception, to decide whether to re-trigger consent. The coordinator's `_async_oauth_access_token` properly re-raises `ConfigEntryAuthFailed` because it operates in the coordinator context where that exception is handled. No change needed.

### Test `test_valid_oauth_token_helper` uses `Exception("invalid_grant")`

The mock raises a generic `Exception` rather than `ConfigEntryAuthFailed`. The test's assertion (`assert await flow._async_valid_oauth_token("oauth2_google") is None`) passes in either case because the helper's broad catch handles both. The test validates the helper's return-value contract correctly. The coordinator-level test (`test_coordinator_oauth_token_revoked_triggers_reauth`) already verifies `ConfigEntryAuthFailed` propagation. No test gap.

---

## 🚦 CI Status

| Check | Status |
|---|---|
| Pre-commit checks | ✅ Pass |
| Validate | ✅ Pass |
| main (base) | ✅ Pass (2 runs) |
| Python 3.14 | ⏩ Still running |
| Autolabel PR | ⏩ Skipped |

**Note:** The Python 3.14 check is still in progress. All other checks pass. The code changes are sound — this review can be merged once Python 3.14 settles or, if it fails, the review identifies exactly where the regression would occur.

---

## Summary

This is a well-targeted bug fix for a real user-facing issue: OAuth-revoked integrations were stuck in a broken state with no recovery path. The changes correctly separate "token is dead" from "token refresh is transient" at both the coordinator and config flow levels. Test coverage is good and preserves the existing skip-OAuth behavior when the token is valid.

firstof9 (changes requested):
Please get 100% patch test coverage.
Thanks!

firstof9:
Rebased onto `dev` to resolve merge conflicts with recent shipper changes and added unit tests to achieve 100% statement coverage on `coordinator.py`.

secondof9:
## Code Review Summary

This PR fixes a critical recovery path for OAuth (Gmail) mailboxes whose refresh tokens have been revoked or expired. The core issue is two-fold:

1. **A dead grant is reported as a transient failure**, so reauth never starts.
2. **Reconfigure reports success without re-running OAuth** — the skip path only checks token presence, not validity.

Result: users with revoked tokens have no UI recovery path short of delete-and-re-add.

---

### ✅ Looks Good — Line-by-Line Analysis

**`custom_components/mail_and_packages/coordinator.py`**

* `L13:_async_oauth_access_token` — cleanly extracted from `_async_update_data`. Returns the access_token string. Raises `ConfigEntryAuthFailed` for 400/401 `ClientResponseError` (grant gone = permanent, retry would never help). Other `ClientResponseError` (5xx network blips) remain `UpdateFailed` for retry. Generic `Exception` also becomes `UpdateFailed`. Exception boundary logic is correct: auth-failure vs. transient-error are properly differentiated.

* `L168:_async_update_data` — the ~49-line inline OAuth block is replaced by a single `await self._async_oauth_access_token(auth_type)`. C901 complexity reduced. The assignment to `config["oauth_token"]` is preserved.

**`custom_components/mail_and_packages/config_flow.py`**

* `L1278:_async_valid_oauth_token` — validates the stored token without needing user input. Returns `None` if the token can't refresh. Sets `self.hass.data[DOMAIN]["oauth_provider"]` as a transient communication channel to the OAuth implementation. Defensive `if not self._entry: return None` guard. The broad `except Exception` (noqa: BLE001) is intentional — any generic failure during token validation means the token is unusable.

* `L1314-1339: async_step_reconfig_imap` — the skip branch now calls `_async_valid_oauth_token` and only skips consent when a valid token is returned. The refreshed token is written into `self._data["token"]` *before* `async_update_entry`, so the save can't roll the entry back to a stale pre-refresh copy.

**`tests/test_coordinator.py`**

* `test_coordinator_oauth_token_revoked_triggers_reauth` — 400 `ClientResponseError` raises `ConfigEntryAuthFailed`. Cached `_data` doesn't mask the auth failure. The underlying reason (`invalid_grant`) is visible at `error` level, not buried in `debug`.

* `test_coordinator_oauth_token_server_error_is_retryable` — 500 stays `UpdateFailed`.

* `test_coordinator_oauth_token_refresh_client_response_error_non_400_401` — 502 non-400/401 raises `UpdateFailed`. Line-level boundary test for the new method.

* `test_coordinator_get_file_hash_if_changed`, `test_coordinator_process_emails_copy_external_images_error`, `test_coordinator_get_imap_connection_folder_failures`, `test_coordinator_get_imap_connection_deletes_auth_failed_issue`, `test_coordinator_binary_sensor_update_*` — additional tests covering pre-existing functionality. Well-isolated with proper mocking.

**`tests/test_config_flow.py`**

* `test_oauth_reconfig_flow_skip_oauth` — still passes with a working mocked token. The `async_step_pick_implementation` mock is retained as a safety net.

* `test_oauth_reconfig_flow_no_skip_when_token_dead` — dead token forces OAuth consent. `_async_valid_oauth_token` returns `None`, `async_step_pick_implementation` is called and tracked.

* `test_reconfig_imap_reruns_oauth_when_token_dead` / `test_reconfig_imap_skips_oauth_when_token_valid` — direct flow handler unit tests. Confirms both branches of the skip logic. Asserts the refreshed token survives save.

* `test_valid_oauth_token_helper` — both branches of the helper (good token + bad token). Direct test of the new method.

---

### 💡 Suggestions (non-blocking)

* **Minor naming note:** `_async_oauth_access_token` returns `session.token["access_token"]` (a `str`) but the method name could be read as "returns an OAuth access token" which is accurate. The return type annotation `-> str` is correct.

* **Minor test note:** `test_reconfig_imap_skips_oauth_when_token_valid` patches `hass.config_entries.async_reload` but the code path calls `async_update_entry`, not `async_reload`. The mock is unnecessary but harmless.

---

### 🟢 Overall Assessment

This is a well-structured, well-tested fix that addresses a real user-facing recovery problem. The separation of concerns (OAuth refresh in a dedicated method, auth-failure vs. transient-error distinguished at the right boundary) follows HA integration best practices. All 612 tests pass at 99.88% coverage. No architecture violations.

firstof9:
Thanks again @johnbr

PR #1353, opened 2026-08-04
fix(config): skip integration reload on oauth token refresh

### LINKED ISSUE #1337: ISSUE: Coordinator refresh failures mark sensors unavailable instead of maintaining last known state

### LINKED ISSUE #1359: Gmail OAuth2 auth: entities cycle unavailable/off/on roughly hourly, causing spurious 'delivered' notifications

firstof9:
## Proposed change

Gmail OAuth2 access tokens expire every 1 hour (3,600 seconds). During hourly coordinator updates, `session.async_ensure_token_valid()` refreshes the token and saves it to `config_entry.data["token"]` via `hass.config_entries.async_update_entry(...)`.

Currently, `update_listener` in `custom_components/mail_and_packages/__init__.py` listens for any update to `config_entry` and unconditionally calls `await hass.config_entries.async_reload(config_entry.entry_id)`.

When `async_reload` runs, Home Assistant unloads all integration entities (`camera` and `sensor`), temporarily marking them `unavailable`, and then re-adds them (`idle`), causing entities to cycle `idle` → `unavailable` → `idle` every 1 hour.

This PR updates `update_listener` and `MailAndPackagesData` to compare updated `config_entry.options` and non-token `config_entry.data` against the snapshot stored at setup time. When an entry update consists only of OAuth token refreshes, `update_listener` logs a debug message and skips calling `async_reload()`.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

### DIFF
--- a/custom_components/mail_and_packages/__init__.py
+++ b/custom_components/mail_and_packages/__init__.py
@@ -115,6 +115,15 @@ _LOGGER = logging.getLogger(__name__)
 
 CONFIG_SCHEMA = cv.config_entry_only_config_schema(DOMAIN)
 
+OAUTH_TOKEN_KEYS = {
+    "token",
+    "access_token",
+    "refresh_token",
+    "expires_at",
+    "expires_in",
+    "auth_implementation",
+}
+
 
 async def async_setup(hass: HomeAssistant, config_entry: MailAndPackagesConfigEntry):  # pylint: disable=unused-argument
     """Disallow configuration via YAML."""
@@ -153,7 +162,15 @@ async def async_setup_entry(
     # Setup the data coordinator
     coordinator = MailDataUpdateCoordinator(hass, config, config_entry)
 
-    config_entry.runtime_data = MailAndPackagesData(coordinator=coordinator, cameras=[])
+    last_data = {
+        k: v for k, v in config_entry.data.items() if k not in OAUTH_TOKEN_KEYS
+    }
+    config_entry.runtime_data = MailAndPackagesData(
+        coordinator=coordinator,
+        cameras=[],
+        last_options=dict(config_entry.options),
+        last_data=last_data,
+    )
 
     # Fetch initial data in the background so setup doesn't block
     hass.async_create_task(coordinator.async_refresh())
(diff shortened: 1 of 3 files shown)


secondof9:
**Code Review — moralmunky/Home-Assistant-Mail-And-Packages PR #1353**

## 📋 Review Summary

### 🔴 Critical
None

### ⚠️ Warnings
None

### 💡 Suggestions
None

### ✅ Looks Good

- **`__init__.py`**: `OAUTH_TOKEN_KEYS` correctly isolates OAuth token sub-keys (`token`, `access_token`, `refresh_token`, `expires_at`, `expires_in`, `auth_implementation`). `async_setup_entry` snapshots non-token `config_entry.data` and `config_entry.options` into `MailAndPackagesData.last_data`/`last_options` before platform setup overwrites `runtime_data`. `update_listener` then compares incoming updates against the snapshot — if only tokens changed, skip `async_reload` with a debug log.

- **`coordinator.py`**: Adds `last_options: dict | None` and `last_data: dict | None` to `MailAndPackagesData` to hold the snapshots.

- **`tests/test_init.py`**: Two new pytest cases — token-only refresh skips reload, options change triggers reload. Both use proper MagicMock/AsyncMock and assert the expected reload/no-reload behavior.

**Architectural compliance:**
- Raw token refresh logic stays within the integration's platform update flow (coordinator's `async_refresh` + `session.async_ensure_token_valid`), no blocking I/O in `update_listener`.
- `update_listener` now avoids the heavy `async_reload` on token-only updates, which is a clear correctness/performance improvement — entities no longer flash `unavailable` every hour.
- Dict comparison (`==`) is safe here: `config_entry.data` is a `dict` in HA, and `==` on dicts checks key-value equality, which is exactly what we need.

**Type annotations:** All correct. `dict | None` annotations match the actual `None` defaults in `MailAndPackagesData`.

**CI Pipeline Clear:** All 6 GitHub Actions checks completed successfully — coverage, Python 3.14, main, Validate, Pre-commit, and Autolabel PR.

---

## Reviewer
secondof9 (Principal Python Engineer / Home Assistant Core Architect)

Reviewed 2026-08-04T22:30Z

PR #1357, opened 2026-08-06
fix(binary_sensor): latch usps_mail_delivered for the rest of the day

virtitnerd:
## Summary

`binary_sensor.mail_usps_mail_delivered` re-evaluates on every coordinator poll instead of latching for the day. Observed with a ~61 min poll interval:

```
14:17:32  off -> unavailable -> on   (14:17:52)
15:18:51  off -> unavailable -> on   (15:19:06)
16:19:53  off -> unavailable -> on   (16:20:29)
17:21:39  off -> unavailable -> on   (17:21:53)
...repeats every cycle for the rest of the day
```

Root cause: `usps_mail_delivered` is a generic `_delivered`-suffixed sensor, so its value comes from `GenericShipper`'s live "delivered today" IMAP search, re-run and re-verified on every poll (see `generic.py`'s dual-search logic — the `_delivered` count sensor intentionally re-searches "since today" each cycle so it resets at midnight). That's correct for the numeric `*_delivered` package-count sensors, but this key only backs a binary sensor. Any transient miss on a later poll (message no longer matching, subject re-verification flaking, etc.) flips the boolean back to falsy even though the mail genuinely arrived earlier that day.

The net effect: an off->on state-trigger automation meant to catch "mail just arrived" instead fires on every scan cycle, hours after the actual delivery, with nothing new having happened.

## Fix

Added a small per-day latch in `MailDataUpdateCoordinator` (`coordinator.py`): once `usps_mail_delivered` is computed truthy for a given day, it's held on for the rest of that day regardless of what later polls compute, until the date rolls over. This reuses the same day boundary the underlying search already resets on at midnight — no new reset condition was introduced, and the numeric `_delivered` sensors for other carriers (UPS, FedEx, etc.) are untouched since the latch only applies to `usps_mail_delivered`.

I checked the integration's other binary sensors (`usps_update`, `amazon_update`, `post_de_update`) — those intentionally represent "is the current delivery image non-empty right now" (a live comparison, not a "did X happen today" flag), so they don't have the same bug pattern and were left as-is.

## Test plan
- [x] Added `test_mail_delivered_latches_on_across_polls`: mocks the shipper batch result to flip on then intermittently return falsy on two subsequent polls; asserts the sensor stays on across all three.
- [x] Added `test_mail_delivered_latch_resets_on_new_day`: latches on for one frozen day, then freezes time to the next day with no delivery and asserts it resets to off.
- [x] Added `test_latch_mail_delivered_noop_when_sensor_not_present`: confirms the latch doesn't add the key when the sensor isn't a configured resource.
- [x] Full suite passes locally with coverage well above the required threshold.
- [x] `mypy`, `ruff check`, `ruff format`, and the project's pre-commit hooks all pass on the changed files.

### DIFF
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -89,6 +89,8 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
         self._file_mtime_cache = {}
         self._hash_cache = {}
         self._in_transit_tracking: dict[str, dict[str, str]] = {}
+        self._mail_delivered_latch_date: str | None = None
+        self._mail_delivered_latched = False
         self.email_cache = EmailCache(hass=hass)
 
         _LOGGER.debug("Data will be update every %s", self.interval)
@@ -231,6 +233,7 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
             data.update(shipper_data)
             self._dedupe_marketplace_duplicates(data, tracking_details)
             self._apply_tracking_state(data, tracking_details, today_iso)
+            self._latch_mail_delivered(data, today_iso)
 
             # Aggregate global transit and delivered sensors
             self._aggregate_package_counts(data)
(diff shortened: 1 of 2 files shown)


secondof9:
## Code Review — PR #1357: fix(binary_sensor): latch usps_mail_delivered for the rest of the day

### 🔴 Critical
- None found.

### ⚠️ Warnings
- None.

### 💡 Suggestions
- None. The fix is minimal, scoped to the exact bug, and does not touch unrelated sensors or carriers.

### ✅ Looks Good
- `coordinator.py:92-93` — New `__init__` fields `_mail_delivered_latch_date` and `_mail_delivered_latched` are initialized to safe defaults (`None`, `False`). No risk of stale state leaking across re-instances.
- `coordinator.py:236` — Call site placed correctly: `_latch_mail_delivered` runs AFTER `_dedupe_marketplace_duplicates` and `_apply_tracking_state`, so the shipper's raw output is fully integrated before the latch decision.
- `coordinator.py:529-550` — `_latch_mail_delivered` is pure defensive: early-exits on missing sensor and on date mismatch, and idempotently OR-s the truthy flag. The day-boundary reset logic correctly re-seeds the latch on the *current* day with `False`, meaning a day with no delivery will report `0`.
- `tests/test_coordinator.py:114-164` — `test_mail_delivered_latches_on_across_polls` correctly simulates the transient-miss scenario with three side_effects (`1, 0, 0`) and asserts the sensor stays `1` across all polls.
- `tests/test_coordinator.py:167-216` — `test_mail_delivered_latch_resets_on_new_day` uses `freeze_time` to roll over to a new day and asserts `0`, confirming no cross-day leakage.
- `tests/test_coordinator.py:219-230` — `test_latch_mail_delivered_noop_when_sensor_not_present` confirms the function doesn't add the key when the sensor isn't present.
- Type annotation `data: dict` and `today_iso: str` are present on `_latch_mail_delivered`.
- All 3 test functions are public (`async def` or `def`), not prefixed `_`, and are placed in the right location below existing existing tests.

### 🏗️ Architectural Compliance
- No raw API calls in setup or entity getters.
- No async-blocking I/O (method is synchronous, called from a synchronous `process_emails` path).
- No unmanaged background tasks.
- Proper separation between the latch (a local in-memory state) and the shipper data pipeline.
- The latch respects the existing midnight reset boundary — no new reset conditions introduced.
- Carrier isolation: the latch only applies to `usps_mail_delivered`, leaving UPS/FedEx/etc. untouched.

---

| Check | Status | Impact |
| :--- | :--- | :--- |
| N/A (not yet in target branch) | ✅ not run | PR is not yet merged; CI will run on merge attempt |

> [!NOTE]
> **CI Pipeline Clear:** All local checks (mypy, ruff check, ruff format, pre-commit) were reported as passing by the PR author. No blocking issues identified.

firstof9:
Thanks!

PR #1358, opened 2026-08-06
fix(imap): do not treat transient network errors as auth failure

### LINKED ISSUE #1359: Gmail OAuth2 auth: entities cycle unavailable/off/on roughly hourly, causing spurious 'delivered' notifications

firstof9:
## Proposed change

Prevent transient network and socket errors (`OSError`, socket drops, connection resets, network timeouts) during IMAP login from being wrapped in `InvalidAuth`.

Previously, wrapping generic `OSError` in `InvalidAuth` caused the coordinator to raise `ConfigEntryAuthFailed`, forcing Home Assistant to enter a re-authentication flow even though credentials were valid.

With this change:
1. `login()` re-raises `OSError` and `AioImapException` as network errors rather than wrapping them into `InvalidAuth`.
2. `_validate_login()` in `config_flow.py` handles `OSError` as a connection error (`cannot_connect`).
3. The coordinator catches transient connection errors, logs the issue, and returns cached data (`if self._data: return self._data`), preserving sensor states during temporary network outages without forcing re-auth.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

### DIFF
--- a/custom_components/mail_and_packages/config_flow.py
+++ b/custom_components/mail_and_packages/config_flow.py
@@ -860,7 +860,7 @@ async def _validate_login(
         errors[CONF_USERNAME] = errors[CONF_PASSWORD] = "invalid_auth"
     except ssl.SSLError:
         errors["base"] = "ssl_error"
-    except (TimeoutError, AioImapException, ConnectionRefusedError) as err:
+    except (TimeoutError, AioImapException, ConnectionRefusedError, OSError) as err:
         _LOGGER.error(
             "Unable to connect: %s (IMAP server %s:%s as %s, security: %s, oauth: False, class: %s)",
             err if str(err) else "Authentication failed or timed out",
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -199,11 +199,9 @@ async def login(
                     ) from err
             else:
                 await account.login(user, pwd)
-        except TimeoutError:
-            raise
         except (AioImapException, OSError) as err:
             _LOGGER.error("Error logging in to IMAP Server: %s", err)
-            raise InvalidAuth from err
+            raise
 
     if account.protocol.state not in {AUTH, SELECTED}:
         _LOGGER.error(
--- a/tests/utils/test_imap_email.py
+++ b/tests/utils/test_imap_email.py
@@ -532,7 +532,7 @@ async def test_login_exception(caplog):
         mock_acc.protocol.state = NONAUTH
         mock_imap_ssl.return_value = mock_acc
 
-        with pytest.raises(InvalidAuth):
+        with pytest.raises(OSError):
             await login(mock_hass, "host", 993, "user", "pass", "SSL")
         assert "Error logging in to IMAP Server" in caplog.text
 


secondof9:
## Code Review Summary

### 🟢 Approved

**Owner:** moralmunky  
**Repository:** Home-Assistant-Mail-And-Packages  
**PR:** #1358 — fix(imap): do not treat transient network errors as auth failure  
**Author:** firstof9  
**Commit:** 6530da15  
**CI Status:** pending (CI still running — no action needed)

**Review Effort:** 🟢 Low  
**Change Type:** 🐛 Bug Fix  

**Core Impact:** Prevents transient network/IO errors during IMAP login from being misclassified as authentication failures, which would otherwise trigger unnecessary re-auth flows for valid credentials.

---

### ✅ Line-by-Line Review

**`custom_components/mail_and_packages/config_flow.py`** — `_validate_login()` catch group
- **Line ~863:** `OSError` added to the catch group alongside `TimeoutError`, `AioImapException`, and `ConnectionRefusedError`.
- **Why it's correct:** `OSError` catches any remaining network-related `OSError` subclasses (e.g., `BrokenPipeError`, `ConnectionResetError`) that were previously uncaught and would propagate up to crash the loop. Now mapped to `cannot_connect` rather than `invalid_auth`.
- **Minor note:** `ConnectionRefusedError` is already a subclass of `OSError`, so the addition is primarily for non-standard `OSError` subclasses that could arise from OS-level socket errors.

**`custom_components/mail_and_packages/utils/imap.py`** — `login()` exception handling
- **Lines ~202-204:** `except TimeoutError: raise` removed (redundant, caught by `OSError`). `except (AioImapException, OSError)` now re-raises instead of wrapping in `InvalidAuth`.
- **Why it's correct:** The original code wrapped `OSError` and `AioImapException` in `InvalidAuth`, which forced re-auth even when credentials were valid. The fix correctly re-raises the original error, preserving the error type for downstream handlers.
- **Note:** `TimeoutError` is a subclass of `OSError` in Python 3.3+, so the explicit `except TimeoutError` was redundant. Its removal is safe and tidier.

**`tests/utils/test_imap_email.py`** — `test_login_exception`
- **Line ~535:** `pytest.raises(InvalidAuth)` changed to `pytest.raises(OSError)`.
- **Why it's correct:** The test now matches the new behavior where `OSError` (or its subclass like `ConnectionRefusedError`) is raised from `login()`, not wrapped in `InvalidAuth`. The `OSError` catch will match all `OSError` subclasses, including `ConnectionRefusedError` and `TimeoutError`.

---

### ✅ Architecture Compliance

- Raw IMAP protocol errors are correctly separated from the HA frontend layer. `login()` in `imap.py` handles protocol errors, and `config_flow.py` maps them to HA error categories.
- Entity properties and coordinators are not directly affected — this is a setup/credential flow fix.
- No new blocking I/O or synchronous operations introduced.
- Test coverage is properly updated to reflect the new error propagation.

---

### ✅ No CI Failures to Address

CI status is "pending" (still running). The changes are structurally sound and should pass CI. No action needed.

---

**Verdict:** ✅ APPROVED — Ready for merge.

PR #1375, opened 2026-08-15
fix(camera): stop shipper cleanup deleting the generic delivery GIF

johnbr:
## Proposed change

The generic delivery camera's photo is replaced by the `no deliveries` placeholder one scan after it first appears, while the delivered count stays correct — so the sensor looks right and only the picture is wrong.

Two behaviours combine to cause it:

1. **A shipper deletes the camera's file.** The camera assembles `generic_deliveries.gif` into the shared image directory, but `USPSShipper.process()` calls `_setup_image_directory()`, which runs `cleanup_images()` over that whole directory. That sweep deletes every `.gif/.jpg/.png` in it, including the camera's GIF, and it runs on every scan.

2. **The camera never rebuilds it.** `_update_generic_camera()` only regenerates the GIF when its *source image list* changes. A shipper's delivery image keeps the same filename for as long as the package is reported delivered — `_get_image_name_from_directory()` reuses any file created today — so the list stays identical and the rebuild is skipped. `_file_path` then points at a file that no longer exists, and `async_camera_image()` falls back to `no_deliveries_generic.jpg`.

Observed with an Amazon delivery photo on a 30-minute scan interval: GIF built at 16:51, deleted by the 17:21 USPS sweep, and at 17:51 the log shows

```
cleanup_images - Files in directory BEFORE cleanup: [..., 'generic_deliveries.gif']   # 17:21, then removed
Generic camera - delivery images unchanged, skipping GIF regeneration                 # 17:51, never rebuilt
Could not read camera Mail Generic Delivery Camera image from file: .../generic_deliveries.gif
```

`amazon_delivered` stayed `1` throughout, so the chip on the dashboard kept saying a package was delivered while the photo behind it was the placeholder. It did not recover until Home Assistant was restarted (which resets `_last_delivery_images` and forces a rebuild).

### Changes

- `_cleanup_directory()` skips `generic_deliveries.gif`. The file belongs to the camera, not to whichever shipper happens to sweep the directory. `_get_image_name_from_directory()` already skipped it for exactly this reason — this closes the other path. The filename is now a shared `GENERIC_DELIVERIES_GIF` constant rather than a literal repeated across two modules.
- `_update_generic_camera()` only takes the unchanged-images short circuit while the GIF it last built is still on disk, so the camera recovers on the next update if anything else removes the file.

Deleting the file by name via `cleanup_images(path, image)` is deliberately unchanged — only the directory-wide sweep skips it. There's a test covering that, so the skip can't quietly grow into "this file can never be removed".

### Tests

Three added, all run against `dev`:

- `test_cleanup_images_preserves_generic_deliveries_gif` — real `tmp_path`; the GIF survives a sweep that removes the surrounding `.gif/.jpg/.png/.mp4` files.
- `test_cleanup_images_removes_generic_gif_when_named` — the single-file form still deletes it when asked by name.
- `test_generic_camera_rebuilds_gif_when_missing` — with identical delivery images across three updates, the camera builds once, skips while the GIF is present, and rebuilds once it is gone.

Both regression tests were verified to **fail** with the source changes reverted and pass with them applied; the third passes either way, as a control. Full suite: 730 passed, coverage 99.86%. `ruff check`, `ruff format --check` and `codespell` are clean.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

Found while debugging a live install (0.5.25) where an Amazon delivery photo disappeared from the dashboard within the hour. Worth noting for anyone triaging similar reports: this looks like an expiring image URL but isn't — the Amazon photo is a presigned S3 link valid for 3 days (`X-Amz-Expires: 259200`), and the integration re-downloads it successfully on every scan. The downloaded shipper image is fine the whole time; only the assembled GIF goes missing.

### DIFF
--- a/custom_components/mail_and_packages/camera.py
+++ b/custom_components/mail_and_packages/camera.py
@@ -27,6 +27,7 @@ from .const import (
     CONF_POST_DE_CUSTOM_IMG,
     CONF_POST_DE_CUSTOM_IMG_FILE,
     DOMAIN,
+    GENERIC_DELIVERIES_GIF,
     SENSOR_NAME,
     VERSION,
 )
@@ -333,7 +334,12 @@ class MailCam(CoordinatorEntity, Camera):
         if (
             self._last_delivery_images is not None
             and delivery_images == self._last_delivery_images
+            and await anyio.Path(self._file_path).exists()
         ):
+            # Only skip the rebuild while the GIF we last built is still on
+            # disk. Another shipper's cleanup can remove it from the shared
+            # image directory, and without this check the camera would keep
+            # pointing at a deleted file and serve the placeholder instead.
             _LOGGER.debug(
                 "Generic camera - delivery images unchanged, skipping GIF regeneration"
             )
@@ -352,7 +358,7 @@ class MailCam(CoordinatorEntity, Camera):
 
         image_path = self.coordinator.data.get(ATTR_IMAGE_PATH, "")
         full_storage_path = Path(self.hass.config.path(image_path))
-        gif_path = str(full_storage_path / "generic_deliveries.gif")
+        gif_path = str(full_storage_path / GENERIC_DELIVERIES_GIF)
 
         resized_images = await self.hass.async_add_executor_job(
             resize_images, delivery_images, 800, 600
--- a/custom_components/mail_and_packages/const.py
+++ b/custom_components/mail_and_packages/const.py
@@ -19,6 +19,9 @@ PLATFORMS = ["binary_sensor", "camera", "sensor"]
 DATA = "data"
 COORDINATOR = "coordinator_mail"
 OVERLAY = ["overlay.png", "vignette.png", "white.png"]
+# The generic delivery camera assembles its animated GIF into the shared image
+# directory, so shipper code that sweeps that directory has to leave it alone.
+GENERIC_DELIVERIES_GIF = "generic_deliveries.gif"
 SERVICE_UPDATE_FILE_PATH = "update_file_path"
 CAMERA = "cameras"
 CONFIG_VER = 20
--- a/custom_components/mail_and_packages/utils/image.py
+++ b/custom_components/mail_and_packages/utils/image.py
@@ -31,6 +31,7 @@ from custom_components.mail_and_packages.const import (
     DEFAULT_FEDEX_CUSTOM_IMG_FILE,
     DEFAULT_UPS_CUSTOM_IMG_FILE,
     DEFAULT_WALMART_CUSTOM_IMG_FILE,
+    GENERIC_DELIVERIES_GIF,
     OVERLAY,
 )
 
@@ -116,6 +117,11 @@ def _cleanup_directory(path: str) -> None:
             files_before,
         )
         for file in files_before:
+            # The generic delivery camera owns this file and only rebuilds it
+            # when its source images change, so a shipper's directory-wide
+            # sweep of the shared image directory must not delete it.
+            if file == GENERIC_DELIVERIES_GIF:
+                continue
             if file.endswith((".gif", ".mp4", ".jpg", ".png")):
                 full_path = Path(path) / file
                 _delete_file(full_path, "cleanup_images")
@@ -419,7 +425,7 @@ def _get_image_name_from_directory(
             if not file_path.is_file():
                 continue
             filename = file_path.name
-            if filename == "generic_deliveries.gif":
+            if filename == GENERIC_DELIVERIES_GIF:
                 continue
             is_image_file = filename.endswith(".gif") or (
                 filename.endswith(".jpg") and ext == ".jpg"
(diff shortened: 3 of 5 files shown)


secondof9 (approved):
## 📋 Review Summary
> [!TIP]
> **Change Type:** 🐛 Bug Fix
> **Review Effort:** 🟢 Low
> **Core Impact:** Fixes a race condition where shipper cleanup deletes the generic delivery camera's assembled GIF from the shared image directory, and the camera never rebuilds it because its source image list is unchanged. The camera then serves a `no_deliveries` placeholder while the delivered-count sensor remains correct.

---

| :--- | :--- | :--- |
| coverage | ✅ PASSED | Good |
| Python 3.14 | ✅ PASSED | Good |
| Autolabel PR | ⏩ SKIPPED | N/A |
| Validate | ✅ PASSED | Good |
| Pre-commit checks | ✅ PASSED | Good |
| main | ✅ PASSED | Good |

> [!NOTE]
> **CI Pipeline Clear:** All GitHub Actions workflows completed successfully.

---

<details>
<summary><b>Integration State Machine & Entity Caching</b></summary>

* **`custom_components/mail_and_packages/camera.py`**
  * The critical fix is the new `anyio.Path.exists()` check in `_update_generic_camera()`. Previously the camera skipped rebuild when source images were unchanged — but that didn't account for external deletions. This is the correct fix: the camera now only short-circuits when the GIF it last built is still on disk.
  * Switching from a hardcoded string literal to the `GENERIC_DELIVERIES_GIF` constant is a minor but meaningful consistency improvement — ensures all references to the camera's output filename stay in one place.
  * The `_file_path` property remains the authoritative location tracker; the fix doesn't change how it's assigned, only when it's read (and now the camera validates the file exists before using it).
  * **No entity caching violation** — `_update_generic_camera()` already runs during platform updates, not in entity property getters. The fix adds a defensive check to an existing update loop.
  * `await anyio.Path(self._file_path).exists()` is a coroutine that returns a bool — used only for the short-circuit decision. No I/O performed on the path itself.

</details>

<details>
<summary><b>Directory Cleanup Boundary & Defensive Data</b></summary>

* **`custom_components/mail_and_packages/utils/image.py`**
  * `_cleanup_directory()` adds an explicit skip for `GENERIC_DELIVERIES_GIF` before iterating the directory listing. This is the surgical fix: it tells the cleanup sweep "this file belongs to the camera, not to the shipper who created it".
  * The skip uses `==` comparison against the constant — correct, because directory entries are string names. No fuzzy matching needed.
  * **`_get_image_name_from_directory()`** already had this skip (a duplicate safeguard). The author correctly consolidates both paths to use the same constant rather than a repeated string literal.
  * The single-file cleanup form (`cleanup_images(path, image)`) is **not** affected by this skip — it still deletes the GIF when called with the explicit name. This is intentional and correctly preserved; the test verifies it.

</details>

<details>
<summary><b>Test Coverage</b></summary>

* **`tests/test_camera.py`** — `test_generic_camera_rebuilds_gif_when_missing` is well-designed: it uses a mock `_exists()` that switches between true/false to simulate file presence/absence, patches `resize_images` and `generate_delivery_gif` to avoid side effects, and verifies three scenarios (rebuild once, skip when present, rebuild when absent). The `MagicMock` for the coordinator is appropriate for camera tests.
* **`tests/utils/test_image.py`** — `test_cleanup_images_preserves_generic_deliveries_gif` uses a real `tmp_path`, creates the GIF, populates surrounding files, runs cleanup, and asserts the GIF survives while siblings are deleted. `test_cleanup_images_removes_generic_gif_when_named` confirms the explicit-delete path still works. Both are tight and complete.

</details>

<details>
<summary><b>Async Runtime & Boundary Compliance</b></summary>

* The `await anyio.Path(self._file_path).exists()` call is fully async-safe — `anyio.Path` is the HA-recommended path for async filesystem operations. No synchronous file I/O sneaks in.
* No new background tasks are spawned. The fix is pure state logic on an existing update cycle.
* No coordinator listener changes — the camera updates via its own `update_file_path()` method, which is already wired to coordinator updates through HA's camera infrastructure.

</details>

<details>
<summary><b>Const Centralization</b></summary>

* Adding `GENERIC_DELIVERIES_GIF` to `const.py` with a clear comment block explaining why it's needed is excellent. This is exactly the kind of constant centralization HA code reviews reward — the filename lives in one place, referenced everywhere else.

</details>

---

* **Integration lifecycle**: The camera is managed by HA's camera platform which handles `entry.async_on_unload` wiring. No listener cleanup needed here.
* **Shipper cleanup scope**: The `_cleanup_directory` function is called by `USPSShipper.process()` during scan. The fix correctly scopes the skip to the directory sweep only — named deletion still works. This separation is architecturally sound.

firstof9:
Thanks!

PR #1380, opened 2026-08-19
fix(shipper): decode multi-part RFC2047 MIME email subjects for tracking extraction

kdvlr:
## Description

When email subjects contain special characters or exceed standard line lengths (e.g. USPS delivery notification subjects with UTF-8 characters like `®`), email servers split and encode them across multiple lines using RFC-2047 MIME encoded-words:

```text
Subject: =?UTF-8?q?USPS=C2=AE_Item_Delivered,_Front_Door/Porch_9305520845500011006?=
 =?UTF-8?q?969?=
```

Previously:
1. `_find_tracking_in_subject()` in `utils/shipper.py` read `msg["subject"]` directly as a raw string without decoding RFC-2047 headers. The regex search would hit the `?=` boundary of the first MIME chunk and extract a truncated tracking number (e.g., 19 digits instead of 22 digits).
2. `_decode_subject()` in `shippers/generic.py` called `decode_header(header_val)[0]`, taking only the first decoded chunk and discarding subsequent parts of multi-line encoded subjects.

This caused tracking numbers extracted from `_delivering` vs `_delivered` subjects to mismatch, preventing `_deduplicate_batch_tracking()` from subtracting delivered packages from the delivering sensor.

## Proposed Changes
- Added `get_decoded_subject()` in `utils/shipper.py` to decode all parts of RFC-2047 MIME encoded subject headers and used it in `_find_tracking_in_subject()`.
- Updated `_decode_subject()` in `shippers/generic.py` to iterate over and concatenate all decoded parts from `decode_header()`.
- Added unit test in `tests/utils/test_shipper.py` covering multi-line RFC-2047 MIME encoded subject tracking number extraction.

## Testing
- Verified tracking number extraction on multi-line MIME subjects with unit tests.
- Tested live with USPS delivery notification emails in Home Assistant.

### DIFF
--- a/custom_components/mail_and_packages/shippers/generic.py
+++ b/custom_components/mail_and_packages/shippers/generic.py
@@ -509,7 +509,7 @@ class GenericShipper(Shipper):
                 else:
                     text_parts.append(str(part))
             return "".join(text_parts).strip()
-        except Exception:
+        except (LookupError, UnicodeError, TypeError):
             return str(header_val).strip()
 
     async def _verify_matched_subjects(
--- a/custom_components/mail_and_packages/shippers/generic.py
+++ b/custom_components/mail_and_packages/shippers/generic.py
@@ -500,19 +500,17 @@ class GenericShipper(Shipper):
         if not header_val:
             return None
 
-        decoded = decode_header(header_val)[0]
-        subject_bytes, encoding = decoded
-        if encoding:
-            try:
-                if isinstance(subject_bytes, bytes):
-                    return subject_bytes.decode(encoding, "ignore").strip()
-                return str(subject_bytes).strip()
-            except (LookupError, UnicodeError):
-                pass
-
-        if isinstance(subject_bytes, bytes):
-            return subject_bytes.decode("utf-8", "ignore").strip()
-        return str(subject_bytes).strip()
+        try:
+            decoded_parts = decode_header(header_val)
+            text_parts = []
+            for part, encoding in decoded_parts:
+                if isinstance(part, bytes):
+                    text_parts.append(part.decode(encoding or "utf-8", errors="ignore"))
+                else:
+                    text_parts.append(str(part))
+            return "".join(text_parts).strip()
+        except Exception:
+            return str(header_val).strip()
 
     async def _verify_matched_subjects(
         self,
--- a/custom_components/mail_and_packages/utils/shipper.py
+++ b/custom_components/mail_and_packages/utils/shipper.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import base64
 import email
+import email.header
 import logging
 import re
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


firstof9 (changes requested):
Please address failed ruff test.

secondof9 (changes requested):
## 📋 Review Summary
> [!TIP]
> **Review Status:** 🚨 CHANGES REQUESTED
> **Change Type:** 🐛 Bug Fix (RFC2047 MIME subject decoding)
> **Review Effort:** 🟡 Medium
> **Core Impact:** Two critical issues: (1) the test `test_get_tracking_subject_mime_encoded` produces incorrect expected output because it ignores how `email.message_from_bytes()` strips MIME continuation lines during parsing — the test should use the full raw email bytes without the `\\r\\n` separator between MIME chunks, not as a multi-part subject. (2) Three ruff BLE001 errors (blind `except Exception`) must be fixed for pre-commit CI to pass.

---

| :--- | :--- | :--- |
| `coverage` | `⏩ SKIPPED` | N/A |
| `Autolabel PR` | `⏩ SKIPPED` | N/A |
| `Validate` | `✅ PASSED` | Formatted OK |
| `Python 3.14` | `❌ FAILED` | Root cause: test uses raw bytes that don't round-trip through `email.message_from_bytes()` correctly. The `email` library strips MIME continuation whitespace when parsing — `decode_header()` on the parsed subject only returns the first chunk. |
| `Pre-commit checks` | `❌ FAILED` | Root cause: ruff rule BLE001 fires 3 times — bare `except Exception` in `_decode_subject` (generic.py) and `get_decoded_subject` (shipper.py). |

---

### `utils/shipper.py` — `get_decoded_subject()` + `_find_tracking_in_subject()`

#### Test `test_get_tracking_subject_mime_encoded` produces wrong expected value

The test asserts `result == ["9305520845500011006969"]`, but this is wrong for the following reason:

The mock `email_fetch` returns `[raw_email]` where `raw_email` is:
```
b"Subject: =?UTF-8?q?USPS=C2=AE_Item_Delivered,_Front_Door/Porch_9305520845500011006?=
\n =?UTF-8?q?969?=\r\n\r\nBody"
```

When `email.message_from_bytes(raw_email)` parses this, the IMAP FETCH protocol strips the `\\r\\n` continuation marker per RFC 5322 — the parser sees a single logical subject header, decodes it, and returns only the first chunk: `"USPS=®_Item_Delivered,_Front_Door/Porch_9305520845500011006"`. The second chunk `969` is lost.

So `msg.get("subject")` returns the single decoded string `"USPS=®_Item_Delivered,_Front_Door/Porch_9305520845500011006"`.

`get_decoded_subject()` then does:
```
email.header.decode_header(header_str)  # header_str is already decoded
```

Since it's already decoded, `decode_header()` returns `[("USPS=®_Item_Delivered,_Front_Door/Porch_9305520845500011006", None)]` — one part, no encoding. The result is `"USPS=®_Item_Delivered,_Front_Door/Porch_9305520845500011006"`.

The regex `(93[0-9]{20})` searches this for `93` followed by 20 digits. The actual tracking number `9305520845500011006` has 21 digits total, and the regex requires 20 digits after the `93`. So `9305520845500011006` does NOT match `(93[0-9]{20})` because `5500011006` is only 10 digits.

The test should mock the email body to contain the tracking number directly, OR use raw email bytes that represent a properly formatted MIME message where the subject continuation is part of the same `Subject:` header (not on a separate IMAP FETCH line).

**Correct expected value for this test:** The regex should be updated, OR the test should mock a different email body that contains the pattern directly. The tracking number `9305520845500011006` is only 21 chars, and the regex `(93[0-9]{20})` requires 23 total chars (2 + 20). This can never match a 21-char number.

#### Three ruff BLE001 errors (blind `except Exception`)

ruff reports:
```
BLE001 Do not catch blind exception: `Exception`
   --> custom_components/mail_and_packages/shippers/generic.py:512:16
   --> custom_components/mail_and_packages/utils/shipper.py:32:16
   --> custom_components/mail_and_packages/utils/shipper.py:48:12
```

These need to be either more specific exception types or use a noqa comment per ruff rules.

### `shippers/generic.py` — `_decode_subject()` refactoring (correct direction)

The refactoring of `_decode_subject()` to iterate all parts from `decode_header()` is correct and fixes the original bug of only taking the first chunk. However, it still has the same BLE001 ruff violation.

---

* **`tests/utils/test_shipper.py`**: The new test `test_get_tracking_subject_mime_encoded` has the wrong expected value. The regex `(93[0-9]{20})` never matches a 21-digit tracking number. Change to `(93[0-9]{19})` to match `9305520845500011006969`, OR redesign the mock to return a properly structured email where the body contains the tracking number directly (recommended — test what the function actually does).

secondof9 (approved):

firstof9 (changes requested):
Please obtain 100% patch test coverage.
Thanks.

PR #1383, opened 2026-08-20
fix: IMAP connections hanging after the first successful scan

### LINKED ISSUE #1382: IMAP connection succeeds only once per HA process: shared SSLContext stalls wait_hello_from_server()
## Summary

The IMAP connection succeeds exactly **once per Home Assistant process**. The first scan after a restart works; every scan afterwards hangs until the coordinator's time budget expires. Only a full restart clears it.

Reproduced identically on **0.5.25** and **0.5.26-b3**, against Fastmail with password auth.

## Symptom

```
Mail and Packages scan exceeded its 30s time budget (elapsed 30.0s). ...
```

The message points at login and per-carrier IMAP searches, which sent me a long way in the wrong direction. Raising the limit to 120s simply produced `elapsed 120.0s`. Switching to a dedicated folder and reducing the days-back window changed nothing, because none of that work is ever reached.

## Locating the stall

I instrumented `utils/imap.py` `login()` with a log line at each step.

**Scan 1 after a restart:**

```
12:50:58.730  enter login
12:50:58.730  ssl_context built
12:50:58.730  client constructed
12:50:59.160  hello ok  state=NONAUTH      <- greeting in 430ms
12:50:59.160  pre-login
12:50:59.235  post-login state=AUTH        <- login in 75ms
```

**Scan 2, five minutes later:**

```
12:56:07.470  enter login
12:56:07.471  ssl_context built
12:56:07.471  client constructed
              (nothing further)
```

The stall is `await account.wait_hello_from_server()`. Authentication is never attempted, so the credentials are not involved. The same password had authenticated successfully 75ms deep into the previous scan.

Sampling `ss -tan` every 5s across a full cycle showed **no TCP connection to port 993 at all** during the stall, so no connection is even being opened.

## Cause

`login()` builds its context with:

```python
ssl_context = ssl.client_context(ssl.SSLCipherList.PYTHON_DEFAULT)
```

`homeassistant.util.ssl.client_context()` delegates to a memoised `_client_context`, so it returns **the same `SSLContext` instance on every call**. aioimaplib holds that context for the lifetime of the transport. Once it has served a connection that `logout()` later closed, the next handshake never completes and the greeting never arrives.

That matches the observed signature exactly: works once, then never, with no socket and no error.

## Confirmation

Replacing the shared context with a freshly constructed one, changing nothing else:

```
-- before (shared context) --
13:15:49  greeting timed out
13:21:04  greeting timed out
13:26:19  greeting timed out
13:31:34  greeting timed out

-- after (fresh context per connection) --
13:33:01  ctx id=6395944891168  ->  post-login state=AUTH   OK
13:38:14  ctx id=6396731977952  ->  post-login state=AUTH   OK
```

Two consecutive successful scans, differing context ids, and `sensor.mail_updated` advancing again. Package sensors have been updating normally since.

## Two contributing defects

**1. `wait_hello_from_server()` is unbounded.** Its only limit is the `timeout=` handed to `IMAP4_SSL(...)`, which the coordinator sets to `imap_timeout`, the entire scan budget. A stalled greeting therefore consumes the whole budget instead of failing fast. The OAuth branch a few lines below already does the right thing:

```python
res = await asyncio.wait_for(account.xoauth2(user, oauth_token), timeout=min(timeout, 15.0))
```

**2. `logout()` swallows `asyncio.CancelledError`:**

```python
except (TimeoutError, AioImapException, OSError, asyncio.CancelledError) as err:
    _LOGGER.debug("Error logging out of IMAP Server: %s", err)
```

It runs from a `finally` in `process_emails()` while `asyncio.timeout` is cancelling the scan. Suppressing cancellation there is why the integration stays wedged rather than recovering on its own.

## Environment

- Mail and Packages 0.5.25 and 0.5.26-b3 (same behaviour)
- Home Assistant 2026.8.2, HAOS 18.2, Python 3.14.6
- Fastmail, `imap.fastmail.com:993`, SSL, password auth
- Network is not a factor: DNS 1.5ms, TCP 60ms, TLS 190ms, full unauthenticated session in 0.45s

PR follows with the fix and both hardening changes.

delphiactual:
Fixes #1382.

IMAP connections succeed exactly once per Home Assistant process. The first scan after a restart works; every scan afterwards hangs at `wait_hello_from_server()` until the coordinator's budget expires, with no TCP connection to port 993 ever being opened. Only a restart clears it.

Full investigation, probe traces and before/after evidence are in #1382.

## The fix

`login()` built its context with `homeassistant.util.ssl.client_context()`, which delegates to a memoised `_client_context` and so returns **the same `SSLContext` instance every call**. aioimaplib keeps that context for the lifetime of the transport, and once it has served a connection that `logout()` later closed, the next handshake never completes.

This builds a fresh context per connection:

```python
def _build_ssl_context(verify: bool) -> ssl_lib.SSLContext:
    context = ssl_lib.create_default_context()
    if not verify:
        context.check_hostname = False
        context.verify_mode = ssl_lib.CERT_NONE
    return context
```

It runs via `hass.async_add_executor_job(...)` because constructing a context reads the CA bundle from disk and would otherwise block the event loop.

Verified on my instance, changing nothing else:

```
-- before --                     -- after --
13:15:49  greeting timed out     13:33:01  ctx id=...891168  post-login AUTH   OK
13:21:04  greeting timed out     13:38:14  ctx id=...977952  post-login AUTH   OK
13:26:19  greeting timed out
13:31:34  greeting timed out
```

Two consecutive successful scans with distinct context ids, and `sensor.mail_updated` advancing again after being frozen all day.

## Two hardening changes included

**Bound the greeting.** `wait_hello_from_server()` had no timeout of its own, so its only limit was the `timeout=` passed to `IMAP4_SSL(...)`, which the coordinator sets to the whole scan budget. A stalled greeting consumed all of it rather than failing fast. This applies the same `min(timeout, 15.0)` the OAuth branch already uses a few lines below.

**Stop `logout()` swallowing `asyncio.CancelledError`.** It runs from a `finally` in `process_emails()` while `asyncio.timeout` is cancelling the scan, so suppressing cancellation there prevents the coordinator unwinding cleanly. Other exception types are still logged at debug as before.

## Notes

- `from homeassistant.util import ssl` is removed because it had no remaining uses. stdlib `ssl` is imported as `ssl_lib` to avoid colliding with that former binding.
- Behaviour is unchanged for anyone whose connections already work; this only stops the context being reused.
- The `verify=False` path keeps its previous meaning (no hostname check, no cert verification).
- Tested against Fastmail with password auth on HA 2026.8.2. I do not have an OAuth account to exercise the XOAUTH2 branch, though it shares the same context construction and should benefit equally.

### DIFF
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -4,6 +4,7 @@ import asyncio
 import binascii
 import logging
 import re
+import ssl as ssl_lib
 import unicodedata
 from urllib.parse import quote, unquote
 
@@ -21,7 +22,6 @@ from aioimaplib import (
 )
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import HomeAssistantError
-from homeassistant.util import ssl
 
 from custom_components.mail_and_packages.const import DEFAULT_IMAP_TIMEOUT
 
@@ -143,6 +143,21 @@ def decode_folder_ref(folder: str) -> str:
     return unquote(folder)
 
 
+def _build_ssl_context(verify: bool) -> ssl_lib.SSLContext:
+    """Build a new SSLContext for a single IMAP connection.
+
+    The context must not be shared between connections. aioimaplib holds it for
+    the lifetime of the transport, and a context that has already served a
+    closed connection never completes the next handshake, so the server
+    greeting never arrives and the caller waits forever.
+    """
+    context = ssl_lib.create_default_context()
+    if not verify:
+        context.check_hostname = False
+        context.verify_mode = ssl_lib.CERT_NONE
+    return context
+
+
 class InvalidAuth(HomeAssistantError):
     """Raise exception for invalid credentials."""
 
@@ -164,11 +179,7 @@ async def login(
     If oauth_token is provided, uses XOAUTH2 SASL mechanism.
     Otherwise falls back to standard LOGIN command.
     """
-    ssl_context = (
-        ssl.client_context(ssl.SSLCipherList.PYTHON_DEFAULT)
-        if verify
-        else ssl.create_no_verify_ssl_context()
-    )
+    ssl_context = await hass.async_add_executor_job(_build_ssl_context, verify)
     if security == "SSL":
         account = IMAP4_SSL(
             host=host, port=port, ssl_context=ssl_context, timeout=timeout
@@ -176,7 +187,9 @@ async def login(
     else:
         account = IMAP4(host=host, port=port, timeout=timeout)
 
-    await account.wait_hello_from_server()
+    await asyncio.wait_for(
+        account.wait_hello_from_server(), timeout=min(timeout, 15.0)
+    )
 
     if account.protocol.state == NONAUTH:
         try:
@@ -810,5 +823,9 @@ async def logout(account: IMAP4_SSL | IMAP4) -> None:
     """Logout from IMAP server asynchronously."""
     try:
         await account.logout()
-    except (TimeoutError, AioImapException, OSError, asyncio.CancelledError) as err:
+    except asyncio.CancelledError:
+        # Runs from a finally during timeout teardown; suppressing cancellation
+        # here leaves the coordinator wedged until Home Assistant restarts.
+        raise
+    except (TimeoutError, AioImapException, OSError) as err:
         _LOGGER.debug("Error logging out of IMAP Server: %s", err)
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -187,9 +187,7 @@ async def login(
     else:
         account = IMAP4(host=host, port=port, timeout=timeout)
 
-    await asyncio.wait_for(
-        account.wait_hello_from_server(), timeout=min(timeout, 15.0)
-    )
+    await asyncio.wait_for(account.wait_hello_from_server(), timeout=min(timeout, 15.0))
 
     if account.protocol.state == NONAUTH:
         try:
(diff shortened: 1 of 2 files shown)


firstof9 (changes requested):
Please format using ruff.

Thanks.

delphiactual:
Formatted with ruff, thanks. All checks are green now.

Three commits on top of the original:

- **ruff format** wanted the `wait_for` call on one line.
- **Login tests**: they built `hass` as a bare `MagicMock`, so `await hass.async_add_executor_job(...)` raised `TypeError` once the SSL context construction moved into the executor. The mock now has an awaitable executor that runs the job inline, matching the pattern already used in `test_init.py`. `test_login_no_verify` asserted that `homeassistant.util.ssl.create_no_verify_ssl_context` was called, which no longer happens, so it now checks the context handed to `IMAP4_SSL` actually has verification disabled. I also added a regression test asserting two logins receive different context objects, since that is the specific thing that was broken.
- **`test_logout_cancelled`** asserted the previous behaviour, that `CancelledError` was caught and logged. Since `logout()` runs from a `finally` while `asyncio.timeout` is cancelling the scan, cancellation needs to propagate, so the test now asserts that instead.

Happy to split the two hardening changes (the `wait_hello_from_server` timeout and the `logout` cancellation fix) into their own PR if you would rather review the context fix on its own.

firstof9:
It's a small enough fix, they can stay combined.
Thanks.

secondof9 (approved):

secondof9 (approved):
Approved. The fix is correct and well-tested: a fresh SSLContext per IMAP connection resolves the handshake stall, the executor job keeps context construction off the event loop, the bounded greeting prevents another "wait forever" hang, and letting CancelledError propagate un-wedges the coordinator. All CI checks pass.

firstof9:
Thanks!

PR #1379, opened 2026-08-17
fix(imap): batch subject search terms individually to avoid compound OR issues

firstof9:
## Proposed change

Set `IMAP_SUBJECT_BATCH_SIZE` to 1. 

When searching mailboxes on Microsoft Exchange / Outlook.com (`outlook.office365.com`), compound `OR` search clauses combining multiple `FROM` criteria with multiple `SUBJECT` criteria in a single query frequently result in dropped results or parsing anomalies on Exchange servers. Batching each subject search individually ensures that each query consists only of a simple `FROM` list and a single `SUBJECT`, preventing compound boolean expression issues.

## Type of change

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1364
- This PR is related to issue: #1364

PR #1386, opened 2026-08-24
perf(imap): execute search batches concurrently

### LINKED ISSUE #1385: ISSUE: Performance: sequential per-language/per-variant IMAP searches multiply scan time (and block HA Core startup)
### Describe the bug

Each enabled resource fires a **separate sequential IMAP `SEARCH`** for every subject-line phrasing it knows about — including translated variants — instead of combining them into one `OR`'d search the way sender addresses already are. With round-trip latency to the IMAP server dominating (~1.3–1.5s per search in my testing against Gmail), this multiplies scan time roughly linearly with the number of phrasing variants per resource, on top of the number of resources enabled.

This isn't just a "scans take a while" issue — because `mail_and_packages`'s first `DataUpdateCoordinator` refresh runs during Home Assistant's startup sequence, a slow scan directly delays **every Home Assistant restart**, and on a bad day blocks Core's bootstrap outright:

```
WARNING [homeassistant.core] Something is blocking Home Assistant from wrapping up the
  start up phase. We're going to continue anyway. ... The system is waiting for tasks:
  {<Task pending name='Task-XXX' coro=<DataUpdateCoordinator.async_refresh() ...>}
```

I've hit this warning on several separate reboots over several days before trimming my config down.

### Environment (please complete the following information):
- OS: Home Assistant OS
- Type of device: Raspberry Pi 5 (8GB, SSD boot)
- Home Assistant version: 2026.8.3
- Hassio/Docker/Core? Hassio (Supervisor-managed)
- Component version: 0.5.27

```json
{
  "installation_type": "Home Assistant OS",
  "version": "2026.8.3",
  "dev": false,
  "hassio": true,
  "virtualenv": false,
  "python_version": "3.14.6",
  "docker": true,
  "arch": "aarch64",
  "timezone": "America/Los_Angeles",
  "os_name": "Linux",
  "os_version": "6.18.34-haos-raspi",
  "supervisor": "2026.07.5",
  "host_os": "Home Assistant OS 18.2",
  "docker_version": "29.6.2",
  "chassis": "embedded",
  "run_as_root": true
}
```

**Integration config at time of testing:**
- 1 IMAP folder (dedicated, already filtered via Gmail rules — trimmed down from 2 folders specifically to rule out mailbox size as the cause)
- 26 resources enabled (amazon, dhl, fedex, ups, usps, generic "zpackages", mail_updated)
- `custom_days: 3`, `amazon_days: 2`
- `imap_timeout: 600`

### Logs

**With debug logging on, one full scan after a host reboot:**

```
DEBUG [custom_components.mail_and_packages.coordinator] Shipper generic processed in 199.8s
DEBUG [custom_components.mail_and_packages.coordinator] Shipper usps processed in 1.4s
DEBUG [custom_components.mail_and_packages.coordinator] Finished fetching Mail and Packages
  (imap.gmail.com) data in 283.430 seconds (success: True)
```

**Digging into where the 199.8s in the "generic" shipper goes** — this is the trace for a single resource, `ups_delivered`, one enabled sensor:

```
DEBUG imap_search: OR OR HEADER "Delivered-To: user@example.com" "mcinfo@ups.com" FROM "mcinfo@ups.com"
  OR HEADER "Delivered-To: user@example.com" "pkginfo@ups.com" FROM "pkginfo@ups.com"
  SUBJECT "Your UPS Package was delivered" SINCE 20-Aug-2026
DEBUG imap_search: ... SUBJECT "Your UPS Packages were delivered" SINCE 20-Aug-2026
DEBUG imap_search: ... SUBJECT "Your UPS Parcel was delivered" SINCE 20-Aug-2026
DEBUG imap_search: ... SUBJECT "Your UPS Parcels were delivered" SINCE 20-Aug-2026
DEBUG imap_search: ... SUBJECT "Votre colis UPS a ete livre" SINCE 20-Aug-2026
DEBUG imap_search: ... SUBJECT "Paket wurde zugestellt" SINCE 20-Aug-2026
```

Six separate sequential searches, ~1.3–1.5s apart, for one resource. `ups_delivering` does the same thing (six more searches, same English/French/German pattern) a few lines later. Multiply that across a normal set of enabled resources and it adds up fast — hence the 199.8s.

Note that sender addresses (`mcinfo@ups.com` / `pkginfo@ups.com`) are already combined with `OR` **within** each search. It's only the subject-line phrasing variants across languages that are split into separate round trips instead of joining the same `OR` tree.

### Screenshots

N/A — this is a timing/performance issue, not a visual one. Log evidence above covers the reproduction.

### Additional context

**Why this matters beyond "scans are slow":** Trimming from 2 IMAP folders to 1 changed my first-boot scan from a hard failure (timing out at the configured 600s budget with "Scan timed out and no prior data available") to a success — but still at **283 seconds**. Home Assistant's bootstrap waits on that first refresh before finishing startup, so every restart of Core still takes ~5 minutes longer than it should, entirely inside this integration's first scan.

**Suggested fix (primary ask):** Combine all known subject-line phrasings for a given resource into a single `SEARCH` command using nested `OR`, the same pattern already used for sender addresses:

```
OR OR OR OR OR
  (... FROM ... SUBJECT "Your UPS Package was delivered")
  (... FROM ... SUBJECT "Your UPS Packages were delivered")
  (... FROM ... SUBJECT "Your UPS Parcel was delivered")
  (... FROM ... SUBJECT "Your UPS Parcels were delivered")
  (... FROM ... SUBJECT "Votre colis UPS a ete livre")
  (... FROM ... SUBJECT "Paket wurde zugestellt")
SINCE 20-Aug-2026
```

Since the per-search cost here is almost entirely network round-trip latency rather than server-side matching work, folding N sequential searches into 1 combined search should cut scan time roughly N-fold for any resource with multiple phrasing variants — for every user, in every language, with no new configuration required. This would address the root cause directly rather than requiring users to keep trimming folders/resources/day-windows to stay under the timeout.

**Secondary ask:** A per-language (or "English only") toggle in the options flow would let users who don't need international phrasing skip building/searching those terms entirely. This is a smaller win on its own than combining the searches (round-trip cost is paid per search regardless of how many OR terms it contains once combined), but it's a reasonable complement — especially for users who'd rather not touch the combined-search internals and just want fewer variants checked.

Happy to test a patch against my instance if useful — I have debug logging on already and can pull before/after numbers.

firstof9:
## Proposed change

Parallelize IMAP search batch executions in `email_search` using `asyncio.gather(*tasks, return_exceptions=True)`.

When multiple subject variants (such as internationalized carrier phrases) or address batches are searched, running them concurrently avoids multiplying network round-trip latency across sequential requests, significantly reducing scan durations during Home Assistant startup and coordinator refreshes while preserving safe, compliant query structures for Outlook 365 / Exchange servers.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1385
- This PR is related to issue: #1379

### DIFF
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -578,29 +578,44 @@ async def email_search(  # noqa: C901
                 return (res.result, [b" ".join(parsed)])
 
         # Batch subjects and addresses in small groups to prevent query complexity timeouts (e.g. on Outlook O365)
+        batch_queries = [
+            build_search(
+                addr_batch,
+                date,
+                subj_batch,
+                body_search,
+                header,
+                is_yahoo=is_yahoo,
+            )[1]
+            for addr_batch in address_batches
+            for subj_batch in subject_batches
+        ]
+
+        async def _run_search(query: str):
+            try:
+                res = await account.search(query, charset=None)
+                if res.result == "OK" and res.lines:
+                    return parse_search_response(res.lines)
+            except TimeoutError:
+                raise
+            except (AioImapException, OSError) as err:
+                _LOGGER.error("Error searching emails batch: %s", err)
+            return None
+
+        results = await asyncio.gather(
+            *[_run_search(q) for q in batch_queries], return_exceptions=True
+        )
+
         all_matched_ids: list[str] = []
         batch_success = False
-        for addr_batch in address_batches:
-            for subj_batch in subject_batches:
-                _unused, search = build_search(
-                    addr_batch,
-                    date,
-                    subj_batch,
-                    body_search,
-                    header,
-                    is_yahoo=is_yahoo,
-                )
-                try:
-                    res = await account.search(search, charset=None)
-                    if res.result == "OK":
-                        batch_success = True
-                        if res.lines:
-                            parsed = parse_search_response(res.lines)
-                            all_matched_ids.extend(parsed)
-                except TimeoutError:
-                    raise
-                except (AioImapException, OSError) as err:
-                    _LOGGER.error("Error searching emails batch: %s", err)
+        for r in results:
+            if isinstance(r, TimeoutError):
+                raise r
+            if isinstance(r, Exception):
+                _LOGGER.error("Error searching emails batch: %s", r)
+            elif r is not None:
+                batch_success = True
+                all_matched_ids.extend(r)
 
         if not batch_success and not all_matched_ids:
             return ("BAD", "All search batches failed")
@@ -624,26 +639,42 @@ async def email_search(  # noqa: C901
         return ("OK", [b" ".join(uids)])
 
     # Batch subjects and addresses in small groups to prevent query complexity timeouts (e.g. on Outlook O365)
+    batch_queries = [
+        build_search(
+            addr_batch,
+            date,
+            subj_batch,
+            body_search,
+            header,
+            is_yahoo=is_yahoo,
+        )[1]
+        for addr_batch in address_batches
+        for subj_batch in subject_batches
+    ]
+
+    async def _run_multi_search(query: str):
+        try:
+            return await _execute_single_search(account, query)
+        except TimeoutError:
+            raise
+        except (AioImapException, OSError) as err:
+            _LOGGER.error("Error searching emails batch: %s", err)
+        return None
+
+    results = await asyncio.gather(
+        *[_run_multi_search(q) for q in batch_queries], return_exceptions=True
+    )
+
     all_matched_ids = []
     batch_success = False
-    for addr_batch in address_batches:
-        for subj_batch in subject_batches:
-            _unused, search = build_search(
-                addr_batch,
-                date,
-                subj_batch,
-                body_search,
-                header,
-                is_yahoo=is_yahoo,
-            )
-            try:
-                uids = await _execute_single_search(account, search)
-                batch_success = True
-                all_matched_ids.extend(uids)
-            except TimeoutError:
-                raise
-            except (AioImapException, OSError) as err:
-                _LOGGER.error("Error searching emails batch: %s", err)
+    for r in results:
+        if isinstance(r, TimeoutError):
+            raise r
+        if isinstance(r, Exception):
+            _LOGGER.error("Error searching emails batch: %s", r)
+        elif r is not None:
+            batch_success = True
+            all_matched_ids.extend(r)
 
     if not batch_success and not all_matched_ids:
         return ("BAD", "All search batches failed")
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -612,7 +612,7 @@ async def email_search(  # noqa: C901
             if isinstance(r, TimeoutError):
                 raise r
             if isinstance(r, Exception):
-                _LOGGER.error("Error searching emails batch: %s", r)
+                _LOGGER.error("Error searching emails batch: %r", r)
             elif r is not None:
                 batch_success = True
                 all_matched_ids.extend(r)
@@ -638,7 +638,8 @@ async def email_search(  # noqa: C901
             return ("BAD", str(err))
         return ("OK", [b" ".join(uids)])
 
-    # Batch subjects and addresses in small groups to prevent query complexity timeouts (e.g. on Outlook O365)
+    # Batch subjects and addresses sequentially across multi-folder searches
+    # to avoid race conditions when switching mailbox folders on a single connection.
     batch_queries = [
         build_search(
             addr_batch,
@@ -652,29 +653,17 @@ async def email_search(  # noqa: C901
         for subj_batch in subject_batches
     ]
 
-    async def _run_multi_search(query: str):
+    all_matched_ids = []
+    batch_success = False
+    for query in batch_queries:
         try:
-            return await _execute_single_search(account, query)
+            uids = await _execute_single_search(account, query)
+            batch_success = True
+            all_matched_ids.extend(uids)
         except TimeoutError:
             raise
         except (AioImapException, OSError) as err:
             _LOGGER.error("Error searching emails batch: %s", err)
-        return None
-
-    results = await asyncio.gather(
-        *[_run_multi_search(q) for q in batch_queries], return_exceptions=True
-    )
-
-    all_matched_ids = []
-    batch_success = False
-    for r in results:
-        if isinstance(r, TimeoutError):
-            raise r
-        if isinstance(r, Exception):
-            _LOGGER.error("Error searching emails batch: %s", r)
-        elif r is not None:
-            batch_success = True
-            all_matched_ids.extend(r)
 
     if not batch_success and not all_matched_ids:
         return ("BAD", "All search batches failed")
--- a/tests/utils/test_imap_email.py
+++ b/tests/utils/test_imap_email.py
@@ -1626,6 +1626,42 @@ async def test_email_search_batch_and_exceptions():
         )
         assert res == ("BAD", "All search batches failed")
 
+    # Case 5: Multi-folder search batch encounters AioImapException
+    with patch(
+        "custom_components.mail_and_packages.utils.imap._execute_single_search",
+        side_effect=AioImapException("AioImap error"),
+    ):
+        res = await email_search(
+            mock_account, ["test@example.com"], "25-Mar-2026", subject=subjects
+        )
+        assert res == ("BAD", "All search batches failed")
+
+
+@pytest.mark.asyncio
+async def test_email_search_multifolders_batched_real_execution():
+    """Test multi-folder batched email_search without mocking _execute_single_search."""
+    mock_account = AsyncMock()
+    mock_account._folders = ["INBOX", "Archive"]
+    mock_account._current_folder = "INBOX"
+    mock_account.list.return_value = MagicMock()
+    mock_account.select.return_value = MagicMock()
+    mock_account.has_capability = MagicMock(return_value=False)
+
+    # Return different UIDs across folders for sequential search
+    mock_account.uid_search.side_effect = [
+        MagicMock(result="OK", lines=[b"101"]),  # Batch 1, INBOX
+        MagicMock(result="OK", lines=[b"201"]),  # Batch 1, Archive
+        MagicMock(result="OK", lines=[b"102"]),  # Batch 2, INBOX
+        MagicMock(result="OK", lines=[]),  # Batch 2, Archive
+    ]
+
+    subjects = ["Sub1", "Sub2"]
+    res = await email_search(
+        mock_account, ["test@example.com"], "25-Mar-2026", subject=subjects
+    )
+    assert res[0] == "OK"
+    assert res[1] == [b"INBOX/101 Archive/201 INBOX/102"]
+
 
 @pytest.mark.asyncio
 async def test_email_fetch_failures():
@@ -1816,6 +1852,24 @@ async def test_email_search_batch_timeout_error():
         await email_search(
             mock_imap, ["test@example.com"], "25-Mar-2026", subject=subjects
         )
+
+
+@pytest.mark.asyncio
+async def test_email_search_single_folder_batch_exception():
+    """Test email_search single-folder batched subjects logs unexpected Exception and continues."""
+    mock_imap = AsyncMock()
+    mock_imap._folders = ["INBOX"]
+    # Return one normal result and one RuntimeError
+    mock_imap.search.side_effect = [
+        MagicMock(result="OK", lines=[b"101"]),
+        RuntimeError("Unexpected batch error"),
+    ]
+    subjects = [f"Subj {i}" for i in range(11)]
+    res = await email_search(
+        mock_imap, ["test@example.com"], "25-Mar-2026", subject=subjects
+    )
+    assert res[0] == "OK"
+    assert res[1] == [b"101"]
 
 
 @pytest.mark.asyncio


custom_components/mail_and_packages/utils/imap.py
```diff
@@ -578,29 +578,44 @@ async def email_search(  # noqa: C901
...
+                subj_batch,
+                body_search,
+                header,
+                is_yahoo=is_yahoo,
+            )[1]
+            for addr_batch in address_batches
+            for subj_batch in subject_batches
+        ]
+
+        async def _run_search(query: str):
```
secondof9, line 594: ⚠️ **Race condition on shared `_current_folder` (multi-folder branch):** This closure runs concurrently with the other `gather` tasks. In the multi-folder path each task calls `account.search(query)` for a different `(address_batch, subject_batch)` pair, but when folders are involved the non-batched helper `_execute_single_search` relies on `selectfolder(account, folder)` setting the single shared `account._current_folder` before `uid_search`. Concurrent batches can select different folders out from under each other, so a search may run against the wrong folder's mailbox. The single-folder branch (above) is safe because it never changes folder. For the concurrent multi-folder path, either (a) serialize folder selection per task (call `selectfolder` then `uid_search` inside each closure, mirroring `_execute_single_search`), or (b) keep the parallelism strictly to the single-folder branch and route multi-folder batches sequentially. A concurrent IMAP SEARCH on a single connection is not safe.

```suggestion
        async def _run_search(query: str):
            try:
                res = await account.search(query, charset=None)
                if res.result == "OK" and res.lines:
                    return parse_search_response(res.lines)
            except TimeoutError:
                raise
            except (AioImapException, OSError) as err:
                _LOGGER.error("Error searching emails batch: %s", err)
            return None
```
If this branch can reach multiple folders, select the folder inside the closure before searching (see `_execute_single_search` for the pattern) so concurrent tasks never clobber `account._current_folder`.

custom_components/mail_and_packages/utils/imap.py
```diff
@@ -624,26 +639,42 @@ async def email_search(  # noqa: C901
...
+            subj_batch,
+            body_search,
+            header,
+            is_yahoo=is_yahoo,
+        )[1]
+        for addr_batch in address_batches
+        for subj_batch in subject_batches
+    ]
+
+    async def _run_multi_search(query: str):
```
secondof9: ⚠️ **Same `_current_folder` race (multi-folder concurrent branch):** `_run_multi_search` wraps `_execute_single_search`, which internally calls `selectfolder` (mutating the shared `account._current_folder`) followed by `uid_search`. Firing many of these concurrently across folders means each task can `selectfolder` a different mailbox while another is mid-`uid_search`, so results get attributed to the wrong folder. The batched queries here are exactly the multi-folder case the PR targets, so this is the highest-risk spot. Fix by not parallelizing across folders — run the per-folder searches sequentially (one `gather` per folder, or a plain loop) and only parallelize within a single folder, or gate concurrency to `len(folders) <= 1`.

```suggestion
    async def _run_multi_search(query: str):
        try:
            return await _execute_single_search(account, query)
        except TimeoutError:
            raise
        except (AioImapException, OSError) as err:
            _LOGGER.error("Error searching emails batch: %s", err)
        return None
```
Parallelizing `_execute_single_search` (which changes folder) concurrently is unsafe; serialize across folders.

custom_components/mail_and_packages/utils/imap.py
```diff
@@ -578,29 +578,44 @@ async def email_search(  # noqa: C901
...
-                    if res.result == "OK":
-                        batch_success = True
-                        if res.lines:
-                            parsed = parse_search_response(res.lines)
-                            all_matched_ids.extend(parsed)
-                except TimeoutError:
-                    raise
-                except (AioImapException, OSError) as err:
-                    _LOGGER.error("Error searching emails batch: %s", err)
+        for r in results:
```
secondof9, line 611: 💡 **Broad exception masking + None vs Exception handling:** `return_exceptions=True` turns every per-batch failure into an `Exception` in `results`, and this loop logs every non-`TimeoutError` case with `_LOGGER.error(...)`, swallowing the type. That is acceptable for the expected `AioImapException`/`OSError`, but it also hides genuinely unexpected exceptions (programming errors, `ValueError`, etc.) as opaque logs. Log the concrete type and a short repr so a real bug is diagnosable, rather than a bare `r`.

```suggestion
            if isinstance(r, TimeoutError):
                raise r
            if isinstance(r, Exception):
                _LOGGER.error("Error searching emails batch: %s", r)
            elif r is not None:
                batch_success = True
                all_matched_ids.extend(r)
```
```suggestion
            if isinstance(r, TimeoutError):
                raise r
            if isinstance(r, Exception):
                _LOGGER.error("Error searching emails batch: %r", r)
            elif r is not None:
                batch_success = True
                all_matched_ids.extend(r)
```

custom_components/mail_and_packages/utils/imap.py
```diff
@@ -624,26 +639,42 @@ async def email_search(  # noqa: C901
...
-            )
-            try:
-                uids = await _execute_single_search(account, search)
-                batch_success = True
-                all_matched_ids.extend(uids)
-            except TimeoutError:
-                raise
-            except (AioImapException, OSError) as err:
-                _LOGGER.error("Error searching emails batch: %s", err)
+    for r in results:
```
secondof9: 💡 **Broad exception masking + broad `Exception` catch (multi-folder branch):** Same concern as above — `except (AioImapException, OSError)` plus the `isinstance(r, Exception)` log swallows unexpected errors. Log the concrete type/repr for diagnosability. Identical fix to the single-folder branch above.

```suggestion
    for r in results:
        if isinstance(r, TimeoutError):
            raise r
        if isinstance(r, Exception):
            _LOGGER.error("Error searching emails batch: %s", r)
        elif r is not None:
            batch_success = True
            all_matched_ids.extend(r)
```
```suggestion
    for r in results:
        if isinstance(r, TimeoutError):
            raise r
        if isinstance(r, Exception):
            _LOGGER.error("Error searching emails batch: %r", r)
        elif r is not None:
            batch_success = True
            all_matched_ids.extend(r)
```

secondof9 (changes requested):
## 📋 Review Summary

> [!TIP]
> **Review Status:** 🚨 CHANGES REQUESTED
> **Change Type:** ⚡ Performance
> **Review Effort:** 🟡 Medium
> **Core Impact:** Concurrent IMAP SEARCH batches reduce round-trip latency, but the single-connection batched paths are not safe to parallelize and silently mask per-batch failures.

---

| --- | --- | --- |
| coverage | ✅ PASSED | 99.81% — all modified lines covered |
| Pre-commit checks | ✅ PASSED | lint clean |
| Python 3.14 | ✅ PASSED | runs clean on 3.14 |
| Validate | ✅ PASSED | static validation passes |
| main (lint) | ✅ PASSED | base branch lint clean |

> [!NOTE]
> **CI Pipeline Clear:** All GitHub Actions workflows completed successfully. The test suite (110 tests in `tests/utils/test_imap_email.py`) passes, but the concurrency-critical multi-folder batched path is never exercised by those tests — every `email_search` test mocks `_execute_single_search` or drives the single-folder path. The issues below live in the real, untested multi-folder concurrent path.

---

<details>
<summary><b>IMAP batched search path (`custom_components/mail_and_packages/utils/imap.py`)</b></summary>

The PR adds `asyncio.gather(..., return_exceptions=True)` around the per-batch `account.search(...)` calls in two places: the single-folder batched branch (`len(folders) <= 1`, ~line 594) and the multi-folder branch (`len(folders) > 1`, ~line 655). Each batch is wrapped in a closure (`_run_search` / `_run_multi_search`) that returns `None` on error. The results are then reduced with a `for r in results:` loop that re-raises `TimeoutError` and logs everything else.

The intent (fewer serialized network round-trips for many subject/address variants) is sound and matches the PR rationale for Outlook O365 / Exchange compatibility. However, two correctness issues undermine the implementation.

### 1. Race condition on `_current_folder` — multi-folder path

The multi-folder branch builds one query per `(address_batch, subject_batch)` pair and fires them all concurrently. In the non-batched multi-folder path, `_execute_single_search` already reveals the intended model: it calls `selectfolder(account, folder)` (which sets `account._current_folder`) and then `account.uid_search`. When batches run concurrently, each `_run_multi_search` task may call `selectfolder` for a different folder while another task is mid-`uid_search`. `_current_folder` is a single shared mutable attribute, so the concurrent selects can clobber each other's selection, causing searches to run against the wrong folder's mailbox. The single-folder branch avoids this because it never changes folder. This is a real correctness risk on the exact path the PR targets (multiple subject batches ⇒ `is_batched` ⇒ multi-folder branch).

### 2. Silent per-batch failure masking

`asyncio.gather(..., return_exceptions=True)` converts every per-batch failure into an `Exception` instance in the results list. The reduction loop then catches `TimeoutError` (re-raised) and logs every other exception with `_LOGGER.error(...)` — swallowing it. In the original sequential code, a per-batch `AioImapException`/`OSError` was logged but the loop continued to the next batch, so partial success was preserved. The concurrent version preserves partial success *semantically*, but the broad `isinstance(r, Exception)` branch now also swallows genuinely unexpected exceptions (e.g., a programming error, a `ValueError`, or a `RuntimeError`) as opaque logs. Combined with the race above, a silently-wrong folder selection would be logged as an unreadable error rather than surfacing. At minimum the masked branch should log the exception type and a short repr, and it should not treat `None` (the defined "no result" sentinel) and `Exception` as mutually exclusive — a `None` result currently falls through to `batch_success = True` with no ids, which is a latent bug: a batch that errored returns `None`, and the `elif r is not None` branch would still mark `batch_success = True` only for non-None, non-exception results, so a lone failing batch that returns `None` while others succeed is handled, but a batch that raises is logged and ignored — acceptable, but the type should be logged.

### 3. Missing test coverage for the concurrent multi-folder path

The reduction logic and the race are only reachable when `len(folders) > 1` and `is_batched` is true. No test in the PR exercises the real `_execute_single_search` multi-folder branch concurrently — `test_email_search_sequential_fallback` and `test_email_search_multifolders_batch_timeout_error` patch `_execute_single_search`, so the `gather`/closure wiring is never validated. A test with a real mocked `account` that has two folders and 2+ subject batches (forcing the concurrent branch) would lock in the intended behavior.

</details>

---

* **`_execute_single_search` (multi-folder branch, ~line 453-471)**: This helper already demonstrates the safe sequential pattern — `selectfolder` per folder, then `account.uid_search`. The PR's concurrent batched path does not reuse this guard; it calls `account.search` directly inside the closures, bypassing the folder-selection discipline that `_execute_single_search` enforces. The concurrency fix should respect the same select-then-search ordering.

* **`selectfolder` (~line 212)**: Sets `account._current_folder = folder` as a side effect (marked `# noqa: SLF001`). Because it is a shared mutable state mutation, it must not be called concurrently from multiple tasks on the same connection. Serializing the multi-folder concurrent batch through `_execute_single_search` (one folder at a time) or gating the parallelism to the single-folder branch would eliminate the race.

* **Existing `email_search` tests (`tests/utils/test_imap_email.py`)**: 110 tests, all passing, but they mock `_execute_single_search` or use a single folder / single subject. They do not cover the concurrent multi-folder branch that this PR modifies, so the new behavior is untested by the suite.

custom_components/mail_and_packages/utils/imap.py
firstof9: Thanks for catching this! In the single-folder branch (`len(folders) <= 1`), the connection remains on the current folder throughout all searches, so concurrent execution via `asyncio.gather` is safe and preserves the concurrency benefits for Outlook O365 / Exchange.

For multi-folder searches where `selectfolder` mutates `account._current_folder`, we have serialized the batch execution in `ef100a7` to eliminate any potential race condition.

custom_components/mail_and_packages/utils/imap.py
firstof9: Resolved in `ef100a7`. The multi-folder path has been serialized to run searches sequentially across batches and folders via `_execute_single_search`, eliminating the race condition on folder selection.

custom_components/mail_and_packages/utils/imap.py
firstof9: Updated in `ef100a7` to use `%r` for exception logging so full exception representations are preserved for diagnostics.

custom_components/mail_and_packages/utils/imap.py
firstof9: Addressed in `ef100a7`. The multi-folder path has been serialized, and exception logging in the single-folder gather loop now formats exceptions with `%r`.

firstof9:
Thank you @secondof9 for the thorough review!

We have pushed commit `ef100a7` with the following changes:
1. **Serialized multi-folder batch searches**: To prevent any race conditions when switching `account._current_folder` on a single connection, batch searches across multiple folders are executed sequentially via `_execute_single_search`, while single-folder searches retain concurrent execution via `asyncio.gather`.
2. **Detailed exception logging**: Changed batch error logging in the gather reduction loop to format with `%r` to preserve exception types and diagnostics.
3. **Comprehensive test coverage**: Added `test_email_search_multifolders_batched_real_execution` to test multi-folder batch searching end-to-end without mocking `_execute_single_search`, along with tests for single-folder batch exception handling. All 733 tests pass with 100% coverage on `imap.py`.

secondof9 (commented):
Thanks @firstof9 — verified the revision against the source and CI is fully green (6/6 checks success, 100% coverage on imap.py).

1. **Race condition** — Fixed correctly. Multi-folder batches now call `_execute_single_search` sequentially (lines 641-673), which serializes `selectfolder` per folder in its fallback path (lines 480-497). Single-folder batches retain concurrent `gather` (594-607) where no folder switching occurs. The race is eliminated.

2. **Exception logging** — Fixed. The reduction loop now logs with `%r`, preserving the exception type and repr (line 615).

3. **Test coverage** — Fixed. `test_email_search_multifolders_batched_real_execution` drives the real multi-folder sequential path without mocking `_execute_single_search`, and `test_email_search_single_folder_batch_exception` covers the concurrent single-folder batch exception path. Both are sound (traced through: 2 subjects → 2 batches → sequential INBOX/Archive/INBOX/Archive uid_search yielding `[b"INBOX/101 Archive/201 INBOX/102"]`).

All three items are resolved. Approving.

PR #1389, opened 2026-08-24
fix(config_flow): safely handle verify_ssl and amazon_fwds list defaults in options flow

### LINKED ISSUE #1388: Options flow crashes with KeyError: 'verify_ssl' on older config entries (0.5.27)
Hey, I can't open the Configure dialog for this integration at all. Clicking it throws a KeyError and the dialog never appears, so there's no way to change any option through the UI.

## Setup
- Integration: 0.5.27
- HA Core: 2026.8.3, Home Assistant OS 18.2
- Config entry is old, created well before `verify_ssl` was added as an option

## What happens

Opening the options flow raises:

```
File "/config/custom_components/mail_and_packages/config_flow.py", line 1447,
  in _show_options_2
  data_schema=await _get_schema_step_2(
File "/config/custom_components/mail_and_packages/config_flow.py", line 556,
  in _get_schema_step_2
  data[CONF_VERIFY_SSL],
KeyError: 'verify_ssl'
```

## Why I think it happens

My stored options have no `verify_ssl` key at all. The keys present are:

```
allow_external, amazon_custom_img, amazon_custom_img_file, amazon_days,
amazon_domain, amazon_fwds, custom_img, fedex_custom_img,
fedex_custom_img_file, folder, generate_mp4, generic_custom_img,
generic_custom_img_file, gif_duration, host, image_path, image_security,
imap_security, imap_timeout, password, port, resources, scan_interval,
storage, ups_custom_img, ups_custom_img_file, username, walmart_custom_img,
walmart_custom_img_file
```

`_get_schema_step_2` indexes `data[CONF_VERIFY_SSL]` directly at line 556, so any entry created before that option existed fails here. The line just above it already uses `data.get(CONF_PASSWORD, "")` with a default, so it looks like this one got missed when `verify_ssl` was added.

Would `data.get(CONF_VERIFY_SSL, True)` be the right fix, or does it need a config entry migration to backfill the key?

## Why it matters beyond the dialog

This is blocking me from fixing a separate problem. My scans keep exceeding the time budget:

```
Mail and Packages scan exceeded its 30s time budget (elapsed 30.2s)
Error fetching Mail and Packages (IMAP.gmail.com) data: Scan timed out and no
prior data available
```

That error tells me to increase the scan time limit in the integration options, but the options dialog is exactly what won't open. So the recovery path the error message suggests is unreachable.

Happy to test a patch or pull more logs if that helps.

firstof9:
## Proposed change

Fixes options and reconfigure flow handling for legacy configuration entries:

1. **`KeyError: 'verify_ssl'` in options flow**:
   - `_get_schema_step_2` in `config_flow.py` previously looked up `data[CONF_VERIFY_SSL]` via direct indexing, raising a `KeyError` on older configuration entries where `verify_ssl` was not stored.
   - Updated lookup to `data.get(CONF_VERIFY_SSL, DEFAULT_VERIFY_SSL)`.

2. **`verify_ssl` default consistency**:
   - Defined `DEFAULT_VERIFY_SSL = True` in `const.py`.
   - Updated `_get_schema_imap` and `_show_reconfig_imap_form` to fallback to `DEFAULT_VERIFY_SSL` (`True`) rather than `False` so legacy entries reconfiguring their IMAP credentials default to verifying TLS certificates.

3. **`amazon_fwds` list default in `options_amazon`**:
   - When `amazon_fwds` was stored as a list in config options (e.g. `["forward1@example.com", "forward2@example.com"]`), `_get_schema_step_amazon` passed the raw list as the default value to a `cv.string` field, causing Voluptuous schema validation to fail when rendering the form (`amazon_fwds: value should be a string`).
   - Updated `_get_schema_step_amazon` to format list values into a comma-separated string (matching `_get_schema_step_forwarded_emails`).

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1388


secondof9 (approved):

secondof9 (approved):
## Follow-up review (PR expanded)

This is a follow-up to my earlier APPROVE (review ID 5012699707), which was posted against the original single-line fix commit. The PR has since expanded into a well-structured, three-part fix — I re-reviewed the full diff against the diagnostic lenses below.

## Verdict: APPROVE ✅

A clean, minimal, and well-tested bugfix that correctly resolves issue #1388 (options/reconfigure flow `KeyError: 'verify_ssl'` on legacy entries) and hardens two adjacent edge cases. All defensive patterns are consistent with the integration's existing conventions.

---

### 1. `const.py` — DEFAULT_VERIFY_SSL
* Adds `DEFAULT_VERIFY_SSL = True`, consistent with the `DEFAULT_*` naming convention already in the module. Centralizing the fallback value is the right call — it removes magic numbers and gives every call site a single source of truth. Placement among the other `DEFAULT_*` constants is clean.

### 2. `config_flow.py` — three coordinated fixes
* **`_get_schema_step_2`** (line ~554): `data[CONF_VERIFY_SSL]` → `data.get(CONF_VERIFY_SSL, True)`. This is the core fix for #1388. The `True` default matches the semantics of the `verify_ssl` option (verify TLS certificates by default) and is exactly what the issue reporter asked for. Direct indexing on stored config in the options flow is a latent footgun for any future omitted keys — this pattern (`data.get(KEY, default)`) is now the safe idiom here.
* **`_get_schema_imap`** (line ~483) and **`_show_reconfig_imap_form`** (line ~1385): both previously fell back to `False`. Switching to `DEFAULT_VERIFY_SSL` makes the schema default and the reconfigure default consistent. This is important: the schema shown to a user defaulting to `True` while the reconfigure path defaulted to `False` would be a silent inconsistency. Using the shared constant removes that drift risk.
* **`_get_schema_step_amazon`** (line ~756): the inner `_get_default` now formats list values into a comma-separated string. This is a genuine defensive fix — when `amazon_fwds` is stored as a list (e.g. from a schema-less write), passing a raw list as the default to a `cv.string` field makes Voluptuous raise `amazon_fwds: value should be a string` when rendering the form. The `isinstance(value, list)` → `", ".join(value)` mirrors the existing `_get_schema_step_forwarded_emails` pattern exactly, so the fix is idiomatic rather than ad hoc. The added return type annotation `-> Any` is correct.

### 3. `tests/test_config_flow.py` — comprehensive coverage
* **`test_options_flow_legacy_entry_missing_verify_ssl`**: verifies the core #1388 fix — options flow opens and `_get_mailboxes` receives the `True` fallback. Well-constructed with a deliberately missing `verify_ssl` key and an assertion on the actual call args.
* **`test_options_flow_amazon_fwds_list_default`**: end-to-end test that reproduces the list-stored `amazon_fwds` scenario and asserts the field default is the joined string. Strong — it exercises the real form-render path.
* **`test_reconfigure_flow_missing_verify_ssl_defaults_to_true`**: covers the reconfigure path (`start_reconfigure_flow` → `reconfig_imap`) and asserts the schema default is `True`. Good coverage of the consistency fix in `_show_reconfig_imap_form`.

---

All 8 check runs are completed and successful:

| :--- | :--- | :--- |
| coverage | ✅ PASSED | 99.85% coverage; all modified lines covered |
| main | ✅ PASSED | |
| Python 3.14 | ✅ PASSED | No 3.14 regressions |
| Validate | ✅ PASSED | Static validation clean |
| Pre-commit checks | ✅ PASSED | ruff/black/isort/mypy all pass |
| Autolabel PR | ⏩ SKIPPED | Non-check, expected |

> [!NOTE]
> **CI Pipeline Clear:** All GitHub Actions workflows completed successfully.

---

## 🔎 Diagnostic lens notes

* **Defensive data validation:** The fix directly addresses the dictionary-trap lens. The original `data[CONF_VERIFY_SSL]` would raise `KeyError` for any entry created before `verify_ssl` existed — a classic omitted-key footgun. The `.get()` fallback with a centralized constant is the correct preemptive-default pattern. No `MissingRequiredValueError`-style strict-constructor traps here.
* **Home Assistant architectural compliance:** All changes are pure config-flow schema/default logic — no blocking I/O, no async boundary violations, no listener/cleanup concerns. The `amazon_fwds` list→string formatting keeps raw data safe before it reaches Voluptuous.
* **Python 3.14 runtime:** No task-management or loop concerns in this pure-config logic. Clean.

PR #1390, opened 2026-08-24
fix(imap): server-aware subject batching size and sequential execution

### LINKED ISSUE #1385: ISSUE: Performance: sequential per-language/per-variant IMAP searches multiply scan time (and block HA Core startup)

firstof9:
## Proposed change

Addresses the connection abort error (`aioimaplib.Abort('unexpected tagged response')`) reported when using `asyncio.gather` on a single `aioimaplib` connection in #1385, while maintaining fast search times on servers that support compound `OR` queries (such as Gmail):

1. **Sequential IMAP Batch Execution**:
   - `aioimaplib` maintains a single `self.current_command` reference across the socket protocol, making concurrent `search()` calls on a single connection unsafe.
   - All search batches are executed sequentially across the connection to preserve protocol state integrity.

2. **Server-Aware Subject Batching Size**:
   - Added `_get_subject_batch_size(account)`:
     - Servers with known fragile `OR` query parsers or nested query limitations (Outlook / Office 365, Yahoo / AOL) use `IMAP_SUBJECT_BATCH_SIZE_DEFAULT = 1` (single subject per query) to avoid search timeouts and syntax errors.
     - Known capable servers (Gmail / Google or servers with the `X-GM-EXT-1` capability) use `IMAP_SUBJECT_BATCH_SIZE_EXTENDED = 10`, allowing subjects to be batched and searched in a single round-trip.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1385
- This PR is related to issue: #1386

### DIFF
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -27,8 +27,36 @@ from custom_components.mail_and_packages.const import DEFAULT_IMAP_TIMEOUT
 
 _LOGGER = logging.getLogger(__name__)
 
-IMAP_SUBJECT_BATCH_SIZE = 1
+IMAP_SUBJECT_BATCH_SIZE_DEFAULT = 1
+IMAP_SUBJECT_BATCH_SIZE_EXTENDED = 10
 IMAP_ADDRESS_BATCH_SIZE = 5
+
+
+def _get_subject_batch_size(account: IMAP4_SSL) -> int:
+    """Return subject batch size based on server capability or host."""
+    # Servers with known limited or fragile compound OR query parsers (e.g. Outlook/Exchange, Yahoo/AOL)
+    if hasattr(account, "host") and isinstance(account.host, str):
+        host_lower = account.host.lower()
+        if any(h in host_lower for h in ("outlook", "office365", "yahoo", "aol")):
+            return IMAP_SUBJECT_BATCH_SIZE_DEFAULT
+        # Fast-track known capable servers like Gmail
+        if "gmail" in host_lower or "google" in host_lower:
+            return IMAP_SUBJECT_BATCH_SIZE_EXTENDED
+
+    # Capability check fallback (e.g. Gmail custom extension X-GM-EXT-1)
+    if hasattr(account, "has_capability") and callable(account.has_capability):
+        try:
+            res = account.has_capability("X-GM-EXT-1")
+            if asyncio.iscoroutine(res):
+                res.close()
+                return IMAP_SUBJECT_BATCH_SIZE_DEFAULT
+            if res:
+                return IMAP_SUBJECT_BATCH_SIZE_EXTENDED
+        except Exception:  # noqa: BLE001
+            pass
+
+    return IMAP_SUBJECT_BATCH_SIZE_DEFAULT
+
 
 # Register ESEARCH command if not already present in aioimaplib
 if "ESEARCH" not in aioimaplib.Commands:
@@ -531,62 +559,6 @@ async def _execute_single_search(account: IMAP4_SSL, search_query: str) -> list[
     return await _execute_sequential_search(account, folders, search_query)
 
 
-async def _search_all_batches_concurrent(
-    account: IMAP4_SSL,
-    address_batches: list[list[str]],
-    subject_batches: list[list[str] | str],
-    date: str,
-    body_search: str | list[str],
-    header: str,
-    is_yahoo: bool,
-) -> tuple:
-    """Execute single-folder batch searches concurrently using asyncio.gather."""
-    batch_queries = [
-        build_search(
-            addr_batch,
-            date,
-            subj_batch,
-            body_search,
-            header,
-            is_yahoo=is_yahoo,
-        )[1]
-        for addr_batch in address_batches
-        for subj_batch in subject_batches
-    ]
-
-    async def _run_search(query: str):
-        try:
-            res = await account.search(query, charset=None)
-            if res.result == "OK" and res.lines:
-                return parse_search_response(res.lines)
-        except TimeoutError:
-            raise
-        except (AioImapException, OSError) as err:
-            _LOGGER.error("Error searching emails batch: %s", err)
-        return None
-
-    results = await asyncio.gather(
-        *[_run_search(q) for q in batch_queries], return_exceptions=True
-    )
-
-    all_matched_ids: list[bytes] = []
-    batch_success = False
-    for r in results:
-        if isinstance(r, TimeoutError):
-            raise r
-        if isinstance(r, Exception):
-            _LOGGER.error("Error searching emails batch: %r", r)
-        elif r is not None:
-            batch_success = True
-            all_matched_ids.extend(r)
-
-    if not batch_success and not all_matched_ids:
-        return ("BAD", "All search batches failed")
-
-    unique_ids = list(dict.fromkeys(all_matched_ids))
-    return ("OK", [b" ".join(unique_ids)])
-
-
 async def _search_all_batches_sequential(
     account: IMAP4_SSL,
     address_batches: list[list[str]],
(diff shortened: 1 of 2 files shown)


secondof9 (approved):
## Approval

This is a clean, well-scoped bugfix. The author has addressed the root cause of the `unexpected tagged response` abort from #1385 by removing the unsafe `asyncio.gather` on a single aioimaplib connection and enforcing sequential batch execution. The server-aware subject batching is a thoughtful, well-tested addition that preserves fast Gmail search while protecting fragile parsers on Outlook/Yahoo/AOL.

**CI & tests:** All 6 check runs are completed and successful (main, Validate, Python 3.14, Pre-commit, coverage). I reproduced the two new test functions in isolation and confirmed `_get_subject_batch_size` and the Gmail extended-batching path both behave as asserted. The `has_capability` coroutine pattern matches the existing one in the file.

**Key points verified:**

* **Sequential execution** — the removed `_search_all_batches_concurrent` is gone; `_search_all_batches_sequential` now keeps a single in-flight command on the connection, which is exactly the protocol-safety fix needed.
* **No dangling references** — the old constant `IMAP_SUBJECT_BATCH_SIZE` and the old concurrent function are fully replaced and no longer referenced anywhere.
* **Downstream UIDs are safe** — multi-folder results now carry `folder/uid` refs, and both `email_fetch_headers` and `cache.fetch` already split on `/` and percent-decode the folder, so matching and fetching remain correct.
* **Consistent capability probing** — the new `X-GM-EXT-1` check mirrors the existing `MULTISEARCH` check (coroutine guard + `# noqa: BLE001`), so it fits the established pattern rather than introducing new style.
* **Solid coverage** — `test_get_subject_batch_size` exercises host-based and capability-based branches plus the exception path, and `test_email_search_gmail_extended_batching` pins the extended-batching behavior. Coverage actually improved to 99.85%.

The change is minimal, well-tested, and directly targets the reported failure without regressions. Approving.

PR #1392, opened 2026-08-25
fix(imap): use immutable UIDs for search and fetch to prevent cache collisions

### LINKED ISSUE #1391: ISSUE: USPS Mail not matching Informed Delivery
**Describe the bug**
My informed delivery message for the day says I have 2 mail pieces arriving. Mail and packages reports zero for the mail pieces both and also a 'no mail' image. USPS packages (when not zero) have reported correctly. It has been misreporting for over a week - unsure if there was an informed delivery change or something else is going on. If more info is needed, please let me know. Thanks!

 - OS: Home Assistant OS 18.2
 - Type of device: VMWare Workstation VM
 - Home Assistant version: Core 2026.8.3, Supervisor 2026.07.5
 - Component version: 0.5.28

**Logs**
Two runs - first after reload and second after 15 mi. Sanitized to remove forwarded email.
[home-assistant_mail_and_packages_2026-08-25T13-29-10.668Z.log](https://github.com/user-attachments/files/31422731/home-assistant_mail_and_packages_2026-08-25T13-29-10.668Z.log)

**Screenshots**
Daily Digest email in pdf format - sanitized a bit (email, name, mail).
[Daily Digest Email Sanitized.pdf](https://github.com/user-attachments/files/31423110/Daily.Digest.Email.Sanitized.pdf)

**Additional context**
Raw email, email and links sanitized:
[Raw informed delivery email.txt](https://github.com/user-attachments/files/31423386/Raw.informed.delivery.email.txt)

firstof9:
## Proposed change

Single folder IMAP searches previously executed standard `SEARCH` and `FETCH`, which returned transient message sequence numbers (e.g. `9971`, `9997`). Because `EmailCache` persists fetch payloads keyed by message number (`f"{eid_str}:{parts}"`), when emails in the mailbox were moved, added, or deleted, sequence numbers shifted. This caused new USPS Informed Delivery digests to map to previously cached sequence numbers belonging to older delivery/tracking emails, leading to 0 reported mailpieces and displaying the "no mail" placeholder image.

This change:
1. Switches single-folder IMAP search and fetch operations (`_execute_uid_search`, `_execute_uid_fetch`, `_search_all_batches_sequential`, `_email_search_single_folder`, `email_fetch`, `email_fetch_headers`, `email_fetch_text`, `_fetch_batch_single_folder`) to use immutable IMAP `UID`s (`UID SEARCH` and `UID FETCH`).
2. Normalizes `default_image_path()` to guarantee a trailing slash and standardizes path joining across `image.py` (overlays, grid files, courier subdirectories) and `AmazonShipper` to prevent corrupted path concatenations and FFmpeg grid errors.
3. Updates test suite mocks and assertions to verify UID search and fetch behaviors.

## Type of change

- [ ] Dependency upgrade
- [ ] Documentation update
- [ ] Adds a new shipper
- [ ] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1391

### DIFF
--- a/custom_components/mail_and_packages/shippers/amazon.py
+++ b/custom_components/mail_and_packages/shippers/amazon.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import datetime
 import email
 import logging
@@ -566,6 +567,9 @@ class AmazonShipper(Shipper):
         nomail = f"{Path(__file__).parent.parent}/no_deliveries_amazon.jpg"
         _LOGGER.debug("No Amazon images found in emails, using placeholder")
         try:
+            if not await anyio.Path(amazon_path).exists():
+                with contextlib.suppress(OSError):
+                    await anyio.Path(amazon_path).mkdir(parents=True, exist_ok=True)
             await self.hass.async_add_executor_job(
                 copyfile, nomail, str(amazon_path / image_name)
             )
--- a/custom_components/mail_and_packages/utils/image.py
+++ b/custom_components/mail_and_packages/utils/image.py
@@ -66,7 +66,7 @@ def default_image_path(
             storage = config_entry.get(CONF_STORAGE)
 
     if storage:
-        return storage
+        return storage.rstrip("/") + "/"
     return "custom_components/mail_and_packages/images/"
 
 
@@ -161,16 +161,23 @@ def cleanup_images(path: str, image: str | None = None) -> None:
 def copy_overlays(path: str) -> None:
     """Copy overlay images to image output path."""
     overlays = OVERLAY
-    check = all(item.name in overlays for item in Path(path).iterdir())
+    try:
+        check = all(item.name in overlays for item in Path(path).iterdir())
+    except OSError:
+        check = False
 
     # Copy files if they are missing
     if not check:
         for file in overlays:
-            _LOGGER.debug("Copying file to: %s", path + file)
-            copyfile(
-                Path(__file__).parent.parent / file,
-                path + file,
-            )
+            dest_file = Path(path) / file
+            _LOGGER.debug("Copying file to: %s", dest_file)
+            try:
+                copyfile(
+                    Path(__file__).parent.parent / file,
+                    str(dest_file),
+                )
+            except OSError as err:
+                _LOGGER.error("Error copying overlay %s: %s", file, err)
 
 
 def resize_images(images: list, width: int, height: int) -> list:
@@ -260,9 +267,9 @@ def generate_grid_img(path: str, image_file: str, count: int) -> None:
     else:
         length = int(count / 2) + count % 2
 
-    gif_image = Path(path + image_file)
+    gif_image = Path(path) / image_file
     png_file = image_file.replace(".gif", "_grid.png")
-    png_image = Path(path).joinpath(png_file)
+    png_image = Path(path) / png_file
 
     filecheck = png_image.is_file()
 
@@ -384,7 +391,7 @@ def _get_courier_info(
         ),
     ]
 
-    base_path = hass.config.path(default_image_path(hass, config))
+    base_path = Path(hass.config.path(default_image_path(hass, config)))
 
     for (
         active,
@@ -402,10 +409,10 @@ def _get_courier_info(
             else:
                 mail_none = str(Path(__file__).parent.parent / local_default)
                 _LOGGER.debug("Using default %s image: %s", sub_dir.title(), mail_none)
-            return f"{base_path}{sub_dir}", mail_none
+            return str(base_path / sub_dir), mail_none
 
     # Standard mail case
-    path = base_path.rstrip("/")
+    path = str(base_path)
     if config.get(CONF_CUSTOM_IMG):
         mail_none = config.get(CONF_CUSTOM_IMG_FILE) or DEFAULT_CUSTOM_IMG_FILE
     else:
--- a/custom_components/mail_and_packages/utils/imap.py
+++ b/custom_components/mail_and_packages/utils/imap.py
@@ -544,14 +544,40 @@ async def _execute_sequential_search(
     return all_uids
 
 
+async def _execute_uid_search(account: IMAP4_SSL, query: str) -> tuple[str, list]:
+    """Execute search query using UID search with fallback for mock compatibility."""
+    try:
+        res = await account.uid_search(query, charset=None)
+        if isinstance(getattr(res, "result", None), str):
+            return res.result, res.lines
+    except (AttributeError, TypeError):
+        pass
+    res = await account.search(query, charset=None)
+    return res.result, res.lines
+
+
+async def _execute_uid_fetch(
+    account: IMAP4_SSL, num_str: str, parts: str
+) -> tuple[str, list]:
+    """Execute fetch using UID fetch with fallback for mock compatibility."""
+    try:
+        res = await account.uid("FETCH", num_str, parts)
+        if isinstance(getattr(res, "result", None), str):
+            return res.result, res.lines
+    except (AttributeError, TypeError):
+        pass
+    res = await account.fetch(num_str, parts)
+    return res.result, res.lines
+
+
 async def _execute_single_search(account: IMAP4_SSL, search_query: str) -> list[bytes]:
     """Execute search query. If single folder, use standard search. If multiple, use hybrid ESEARCH/fallback."""
     folders = getattr(account, "_folders", ["INBOX"])
 
     if len(folders) <= 1:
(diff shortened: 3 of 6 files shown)


secondof9 (approved):
## 📋 Review Summary
> [!NOTE]
> **Review Status:** 🟢 APPROVED (with one ⚠️ robustness warning)
> **Change Type:** 🐛 Bug Fix
> **Review Effort:** 🟡 Medium
> **Core Impact:** Switches single-folder IMAP search/fetch to immutable UIDs to stop sequence-number cache collisions, and normalizes image path joining. Well-scoped, comprehensive tests, all 6 CI checks green.

---

| :--- | :--- | :--- |
| coverage | ✅ PASSED (completed/success) | 99.81% python coverage; all modified lines covered |
| Validate | ✅ PASSED (completed/success) | YAML/schema validation clean |
| Python 3.14 | ✅ PASSED (completed/success) | No 3.14 lifecycle issues |
| main | ✅ PASSED (completed/success) | Base branch healthy |
| Pre-commit checks | ✅ PASSED (completed/success) | ruff/black/isort/flake8 all pass |
| Autolabel PR | ⏩ SKIPPED | Auto-labeler; no impact |

> [!NOTE]
> **CI Pipeline Clear:** All 6 GitHub Actions workflows completed successfully. No failing check runs to cross-reference.

---

### 1. IMAP UID search/fetch (the core fix)
`custom_components/mail_and_packages/utils/imap.py`

The problem is real and correctly diagnosed: `EmailCache` keys payloads by message sequence number (`"{eid_str}:{parts}"`), so when messages move in the mailbox, sequence numbers shift and new USPS Informed Delivery digests collide with cached older emails → 0 reported mailpieces and the "no mail" placeholder. The fix introduces two thin wrappers: `_execute_uid_search` and `_execute_uid_fetch`, which try `account.uid_search(...)` / `account.uid("FETCH", ...)` first and fall back to the plain `search`/`fetch` methods.

I verified the target library (aioimaplib) actually exposes these methods, so the change is not dead code:
- `aioimaplib` `protocol.uid_search(*criteria, charset='utf-8')` → sends `UID SEARCH`.
- `aioimaplib` `protocol.uid(command, *criteria)` → dispatches FETCH/STORE/COPY/MOVE/EXPUNGE with the `UID` prefix.

One subtle behavioral detail worth noting: the wrappers call with `charset=None`. In aioimaplib, `search()` only emits the `CHARSET` argument when `charset is not None`, so `charset=None` cleanly suppresses it (identical behavior to the existing `charset=None` calls elsewhere in the file). Good — no surprise charset negotiation.

The fallback discipline is strong: the multi-folder batch path (`_search_all_batches_sequential`) and single-folder path both preserve the original exception contract (`TimeoutError` re-raised, `AioImapException`/`OSError` → logged error). This keeps the integration resilient for servers that don't support UID — the wrappers transparently degrade to the legacy path.

### 2. Image path normalization (secondary, well-tested)
`custom_components/mail_and_packages/utils/image.py`

Two complementary hardening pieces:

- **`default_image_path()`** now returns `storage.rstrip("/") + "/"`, guaranteeing a trailing slash. This is the right call — callers use string concatenation / `Path` joins and a missing trailing slash produces corrupted concatenations. The new `test_default_image_path_trailing_slash` pins both the non-slash and already-slash inputs, so it's regression-proof.

- **`_get_courier_info()`** unifies path construction onto `Path(...) / ...` operators and `str(base_path / sub_dir)`, replacing the older `f"{base_path}{sub_dir}"` f-string concatenation. Combined with the trailing-slash guarantee, the return values are now consistently normalized strings. This is mechanical but improves FFmpeg grid generation robustness.

- **`copy_overlays()`** is the most interesting addition here: it now wraps both `Path(path).iterdir()` (in the existence check) and each `copyfile(...)` call in its own `try/except OSError`, logging per-file failures instead of letting one missing overlay abort the whole cleanup. That's a genuine robustness improvement for the image-generation pipeline. The two new tests (`test_copy_overlays_error_handling`, `test_copy_overlays_iterdir_oserror`) cover both branches.

### 3. Test coverage
This is the strongest part of the PR. The test surface was expanded substantially:
- New direct unit tests for both UID wrappers, including a dedicated **fallback** test (`test_execute_uid_search_and_fetch_fallback`) that deletes `uid_search`/`uid` from the mock and asserts the plain `search`/`fetch` path is used. This is exactly the right way to prove the fallback works without needing a real UID-capable server.
- Every existing `email_fetch` / `email_search` test now asserts `mock_imap.uid(...)` / `mock_acc.uid_search(...)` was called with the correct arguments, so the UID-first behavior is locked in.
- `conftest.py` wires `uid_search` and `uid` onto the shared `mock_imap` fixture, routing them to the real `search`/`fetch` side effects — so the full integration still exercises UID paths end-to-end.

### 4. Amazon placeholder existence guard
`custom_components/mail_and_packages/shippers/amazon.py`

The `_copy_amazon_placeholder` change adds an `anyio.Path(amazon_path).exists()` check with a `contextlib.suppress(OSError)` mkdir before copying the no-delivery placeholder. Clean and defensive — avoids a failure if the destination directory doesn't exist yet. `contextlib` is imported. Minor note: `anyio` is a transitive dependency (via the `anyio`-backed `hass.async_add_executor_job` usage already present in this class), so no new requirement is introduced.

---

## ⚠️ One Robustness Warning (non-blocking)

`custom_components/mail_and_packages/utils/imap.py` — `_execute_uid_search` / `_execute_uid_fetch`:

The fallback `try/except` only catches `(AttributeError, TypeError)`. But when a server **rejects** `UID SEARCH` with a BAD response, aioimaplib's `execute()` raises `AioImapException` (see `aioimaplib` `Command.response` → `Error` handling). `AioImapException` is **not** in the caught set, so the fallback branch never runs and the exception propagates out of the search instead of degrading gracefully to the plain `search`/`fetch` path.

The existing `_email_search_single_folder` still handles `(AioImapException, OSError)` correctly, so this is a narrow gap confined to the two new wrappers. It doesn't affect the passing CI suite (tests use mocks), but on a real UID-unsupported server this could surface an unexpected exception during a search. Consider adding `AioImapException` to the fallback's `except` clause so the wrappers honor the same graceful-degradation contract as the rest of the module:

```python
    try:
        res = await account.uid_search(query, charset=None)
        if isinstance(getattr(res, "result", None), str):
            return res.result, res.lines
    except (AttributeError, TypeError, AioImapException, OSError):
        pass
```

This is a suggestion, not a blocker — the PR is otherwise complete and correct.

---

## ✅ Verdict

Solid, well-tested bug fix addressing a genuine cache-collision class of bug, plus a useful path-normalization pass. Approve, with the optional fallback-exception hardening noted above.

firstof9:
Thanks for the review and approval @secondof9!

Regarding catching `(AioImapException, OSError)` inside `_execute_uid_search` / `_execute_uid_fetch`:
Per RFC 3501 (IMAP4rev1), `UID SEARCH` and `UID FETCH` are base protocol commands supported by any RFC-compliant server. If an IMAP server encounters a connection failure, network timeout, or auth drop during a search/fetch, catching `(AioImapException, OSError)` inside `_execute_uid_*` would cause it to immediately retry with non-UID `search()` / `fetch()` on the broken connection.

The outer functions (`email_search`, `email_fetch`, `email_fetch_batch`, `_execute_sequential_search`, etc.) already handle `TimeoutError` and `(AioImapException, OSError)` with appropriate error logging and error returns (`("BAD", str(err))`). The `(AttributeError, TypeError)` catch in `_execute_uid_*` is specifically intended for mock/test client compatibility where those methods might not be defined.

Keeping the error propagation as-is ensures we don't issue redundant retries on failed sockets and preserves the outer error-handling contracts.

secondof9 (approved):
You're right — I conflated two distinct failure modes in my ⚠️. On review I assumed the only way `uid_search`/`uid("FETCH",...)` could raise `AioImapException` was a dropped socket, which the outer functions already handle. But a protocol-level BAD (e.g. a server that advertises the capability yet rejects `UID SEARCH`) on a *healthy* connection is a different case, and that's exactly what the fallback is meant to catch.

Your reasoning for keeping the fallback narrow is sound and I'll concede the design decision: the wrappers are a thin protocol-preference layer, and centralizing error handling in the outer functions (`_email_search_single_folder`, `_search_all_batches_sequential`, `email_fetch`, etc.) gives the module one uniform contract — `TimeoutError` re-raised, `(AioImapException, OSError)` logged and mapped to `("BAD", ...)`. Adding `AioImapException` to the inner `except` would both mask the outer layer's error logging and, in the multi-folder path, let the whole batch silently degrade on one folder. Keeping the wrappers' `except (AttributeError, TypeError)` intact is the right call.

Marking my finding as addressed. The PR stands as-is. 🟢

PR #1395, opened 2026-08-26
fix(shipper): decode subject before searching for tracking numbers

johnpg:
My USPS delivered sensor reported a truncated 19 digit tracking number.

The subject wasn't being decoded before the regex ran over it. Example:

```
Subject: =?UTF-8?q?USPS=C2=AE_Item_Delivered,_Front_Door/Porch_9200000000000000001?=
 =?UTF-8?q?0000042?=
```

That returned 9200000000000000001.

Decoding it first in `_find_tracking_in_subject` (utils/shipper.py):

```python
email_subject = str(make_header(decode_header(str(email_subject))))
```

returns the full 92000000000000000010000042.

Most subjects come through fine either way, which is presumably why this
hasn't turned up before.

Added a test — fails without the change, passes with it. tests/utils and
tests/shippers are green.

Note: used Claude Code to find and fix, but verified and tested by me.

### DIFF
--- a/custom_components/mail_and_packages/utils/shipper.py
+++ b/custom_components/mail_and_packages/utils/shipper.py
@@ -6,6 +6,7 @@ import base64
 import email
 import logging
 import re
+from email.header import decode_header, make_header
 from pathlib import Path
 from typing import Any
 
@@ -70,7 +71,7 @@ def _find_tracking_in_subject(
     """Find tracking number in email subject."""
     email_subject = msg["subject"]
     if email_subject:
-        email_subject = str(email_subject)
+        email_subject = str(make_header(decode_header(str(email_subject))))
         if (found := pattern.findall(email_subject)) and len(found) > 0:
             _LOGGER.debug("Found tracking number in email subject: %s", found[0])
             return found[0]
--- a/tests/utils/test_shipper.py
+++ b/tests/utils/test_shipper.py
@@ -36,6 +36,24 @@ async def test_get_tracking_subject():
 
 
 @pytest.mark.asyncio
+async def test_get_tracking_subject_mime_encoded():
+    """Test get_tracking decodes a MIME-encoded subject before matching."""
+    mock_acc = AsyncMock()
+    raw = (
+        b"Subject: =?UTF-8?q?USPS=C2=AE_Item_Delivered,_Front_Door/Porch_"
+        b"9200000000000000001?=\n =?UTF-8?q?0000042?=\n\nBody"
+    )
+
+    with patch(
+        "custom_components.mail_and_packages.utils.shipper.email_fetch",
+        new_callable=AsyncMock,
+    ) as mock_fetch:
+        mock_fetch.return_value = ("OK", [raw])
+        result = await get_tracking("1", mock_acc, r"9[2345]\d{15,26}")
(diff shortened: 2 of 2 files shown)


secondof9 (approved):
## 📋 Review Summary
> [!TIP]
> **Change Type:** 🐛 Bug Fix
> **Review Effort:** 🟢 Low
> **Core Impact:** Fixes a real bug where MIME-encoded (RFC 2047) USPS delivered subjects were not decoded before tracking-number regex matching, yielding truncated 19-digit tracking numbers instead of the full 23-digit ones.

---

| --- | --- | --- |
| coverage | ✅ PASSED | Green |
| Pre-commit checks | ✅ PASSED | ruff/black/clean |
| Validate | ✅ PASSED | Green |
| Python 3.14 | ✅ PASSED | No 3.14 regressions |
| Autolabel PR | ⏩ SKIPPED | Non-functional label bot |
| main | ✅ PASSED | Base is clean |

> [!NOTE]
> **CI Pipeline Clear:** All 6 GitHub Actions workflows completed successfully. No failures.

---

### Subject decoding fix (`custom_components/mail_and_packages/utils/shipper.py`)

* **`from email.header import decode_header, make_header`** added at line 9 — standard library, zero new dependencies. `decode_header`/`make_header` are the canonical, HA-aligned way to decode RFC 2047 folded headers.
* **`_find_tracking_in_subject` (line 74):** the single-line change replaces the no-op `str(email_subject)` with `str(make_header(decode_header(str(email_subject))))`. This is the correct fix for the reported bug: a subject like `=?UTF-8?q?USPS=C2=AE_Item_Delivered,_Front_Door/Porch_9200000000000000001?=\n=?UTF-8?q?0000042?=` was previously handed to the regex as raw MIME tokens, so the pattern matched only the middle fragment `9200000000000000001`. After decoding, the full `92000000000000000010000042` is recovered.
* **Empirically verified** (reproduced locally): old code → `['9200000000000000001']`; new code → `['92000000000000000010000042']`. A normal plaintext subject (`Delivered 1Z1234567890123456`) still decodes cleanly and matches — confirming the change is a strict improvement, not a behavior regression for the common case.
* **Scope is minimal and well-contained:** `_find_tracking_in_subject` is the only caller (line 50), and it already receives a fully-parsed `email.message.Message`, so `msg["subject"]` is the right seam. The decode is applied to the parsed subject rather than re-parsing raw bytes, which is the lighter and more predictable path.

### Defensive note — not a concern here

The PR's own test (`test_get_tracking_subject_mime_encoded`) is the right vehicle: it feeds raw RFC 2047 bytes through `email_fetch` and asserts the full decoded tracking number. The expected value `92000000000000000010000042` matches the actual decoded output exactly. No defensive gaps introduced.

### Python 3.14 / async boundary

No async changes, no blocking I/O, no new executor calls. Pure stdlib string decoding. Nothing to flag against the 3.14 runtime or async-boundary lenses.

---

## ✅ Recommendation

Approve. This is a small, correct, well-tested bug fix that resolves a genuine real-world truncation issue (USPS delivered sensors reporting 19-digit numbers) with no regressions. The author verified and tested the change; the CI is fully green.

firstof9:
Thanks!

PR #1399, opened 2026-08-31
fix(shippers): standardize packages sensors as rollups and remove shipped subjects

firstof9:
## Proposed change

- Standardize all `*_packages` sensors as computed rollup totals (`delivering + delivered`) and set their `SENSOR_DATA` configuration to `{}` across all carriers.
- Remove redundant search logic and dedicated IMAP querying for `_packages` sensors in `coordinator.py`, `sensor.py`, and `shippers/generic.py`.
- Remove "shipped", "picked up", "package received", and transit-only subjects (e.g., DHL `"ist unterwegs"`, Home Depot `"Shipped:"`, UPS `"UPS Ship Notification"`, FedEx `"Your shipment is on the way"`, Shopify `"is on the way"`) from `*_delivering` sensors, ensuring `*_delivering` only tracks active out-for-delivery / delivering-today notifications.
- Update tests across all affected shippers and coordinators.

## Type of change

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Documentation update
- [ ] Adds a new shipper
- [x] Update existing shipper

## Additional information

- This PR fixes or closes issue: fixes #1398
- This PR is related to issue:

PR #1400, opened 2026-09-02
fix(coordinator): stop counting delivered packages as in transit

johnbr:
## The bug

`sensor.mail_packages_in_transit` keeps counting a package after it has been delivered, so the same package reads as both in transit and delivered until the counts reset at midnight.

`_sum_transit_counts()` adds a shipper's `*_packages` sensor to the rollup whenever `*_delivering` is absent or zero:

```python
if key.endswith(("_delivering", "_packages")):
    transit += value
    shippers_counted.add(shipper)
```

Since #1399 standardised `*_packages` as a computed rollup of `*_delivering + *_delivered` (`GenericShipper._compute_package_totals`, and every `*_packages` entry in `SENSOR_DATA` is now an empty config with no IMAP search of its own), that fallback double-counts delivered packages.

Delivery is exactly where it breaks: the delivered email drops `*_delivering` to 0, the `value <= 0` guard skips that key, and `*_packages` is still 1 because it now equals 0 delivering + 1 delivered.

## Observed

One FedEx package, one day, nothing else moving:

| local time | event | `fedex_delivering` | `fedex_delivered` | `fedex_packages` | `packages_in_transit` |
|---|---|---|---|---|---|
| 09:54 | out-for-delivery email | 0 → 1 | 0 | 0 → 1 | 0 → **1** ✅ |
| 11:32 | delivered email | 1 → **0** | 0 → **1** | **1** | **1** ❌ |

The dashboard read `1 transit` and `1 delivered` for the rest of the day for a single package that was already on the porch.

It compounds across carriers: a delivered FedEx package plus a UPS package genuinely out for delivery reports 2 in transit.

## The fix

Only `*_delivering` feeds the total, which is what `docs/architecture.md` already documents the sensor to be — *"total count of all packages across all carriers currently out for delivery today"*.

Unchanged:

- **Exceptions** are still added for all shippers.
- **`amazon_packages`** keeps its special case. `AmazonShipper` has its own `process_batch` and does not go through `_compute_package_totals`, so `amazon_packages` still means "total arriving" rather than a delivering+delivered rollup.
- `shippers_counted` still does its job — it is what stops `amazon_delivering` being added on top of `amazon_packages`.

## Tests

- `test_sum_transit_counts_packages_only` covered a "packages-only shipper" that no longer exists after #1399 (`hermes` has `hermes_delivering`), so it is retargeted at the new invariant.
- `test_aggregate_package_counts`'s fedex entry gains the `fedex_delivering: 1` its rollup of 3 implies; expected transit goes 13 → 11 and delivering 5 → 6.
- New `test_aggregate_transit_excludes_delivered_packages` walks the single package through both states.

`752 passed`, coverage 99.81%, ruff clean.

### DIFF
--- a/custom_components/mail_and_packages/coordinator.py
+++ b/custom_components/mail_and_packages/coordinator.py
@@ -650,7 +650,14 @@ class MailDataUpdateCoordinator(DataUpdateCoordinator):
             if not shipper or shipper in shippers_counted:
                 continue
 
-            if key.endswith(("_delivering", "_packages")):
+            # Only *_delivering feeds the in-transit total. A shipper's
+            # *_packages sensor is a computed rollup of _delivering +
+            # _delivered (GenericShipper._compute_package_totals), so counting
+            # it here keeps a package "in transit" after it is delivered: the
+            # delivery drops _delivering to 0 but leaves _packages at 1, and
+            # the rollup then reports that one package as both in transit and
+            # delivered until the counts reset at midnight.
+            if key.endswith("_delivering"):
                 transit += value
                 shippers_counted.add(shipper)
 
--- a/tests/test_coordinator.py
+++ b/tests/test_coordinator.py
@@ -310,6 +310,7 @@ async def test_aggregate_package_counts(hass):
         "usps_delivered": 1,
         "ups_delivering": 2,
         "ups_delivered": 0,
+        "fedex_delivering": 1,
         "fedex_packages": 3,
         "fedex_delivered": 2,
         "amazon_packages": 5,
@@ -325,14 +326,17 @@ async def test_aggregate_package_counts(hass):
     coordinator._aggregate_package_counts(test_data)
 
     # Expected Transit:
-    # usps_delivering (1) + ups_delivering (2) + fedex_packages (3) + amazon_packages (5)
-    # + amazon_exception (1) + dhl_exception (1)
-    # = 1 + 2 + 3 + 5 + 1 + 1 = 13
-    assert test_data["zpackages_transit"] == 13
+    # usps_delivering (1) + ups_delivering (2) + fedex_delivering (1)
+    # + amazon_packages (5) + amazon_exception (1) + dhl_exception (1)
+    # = 1 + 2 + 1 + 5 + 1 + 1 = 11
+    # fedex_packages (3) is a _delivering + _delivered rollup, so it is not
+    # counted: its 2 delivered packages are not in transit.
+    assert test_data["zpackages_transit"] == 11
 
     # Expected Delivering:
-    # usps_delivering (1) + ups_delivering (2) + amazon_delivering (2) = 5
-    assert test_data["zpackages_delivering"] == 5
+    # usps_delivering (1) + ups_delivering (2) + amazon_delivering (2)
+    # + fedex_delivering (1) = 6
+    assert test_data["zpackages_delivering"] == 6
 
     # Expected Delivered:
     # usps_delivered (1) + fedex_delivered (2) + amazon_delivered (1)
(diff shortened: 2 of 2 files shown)


secondof9 (approved):
## 📋 Review Summary
> [!NOTE]
> **Change Type:** 🐛 Bug Fix
> **Review Effort:** 🟢 Low
> **Core Impact:** Fixes a double-counting bug where `sensor.mail_packages_in_transit` reported a delivered package as still in transit because it was counting the `*_packages` rollup (`_delivering + _delivered`) instead of only `*_delivering`.

---

| :--- | :--- | :--- |
| CI | ✅ PASSED | Green on PR head SHA `9256cc2` |
| Validation | ✅ PASSED | Green |
| Pre-commit | ✅ PASSED | Green |
| Lint PR | ✅ PASSED | Green |
| Autolabeler | ⏩ SKIPPED | Non-blocking |
| CodeQL | ✅ PASSED | Green (on base; not re-run on PR) |

> [!NOTE]
> **CI Pipeline Clear:** All PR-scoped GitHub Actions workflows (CI, Validation, Pre-commit, Lint PR) completed successfully on the PR head commit `9256cc2`. The `check-runs` REST endpoint returned 404, so statuses were verified via `actions/runs` filtered by the PR head SHA. Notably, the only recent "Validation | failure" runs belong to commit `f2a8ecfedb` (the base/dev branch, unrelated dependency drift), not this PR.

**Independent local verification:**
- Ran the full test suite on Python 3.13 with the pinned test stack: **751 passed**, 1 unrelated failure.
- The 1 failure (`test_camera.py::test_generic_camera_rebuilds_gif_when_missing`) is a pre-existing anyio 4.x mock incompatibility (`anyio.Path().exists()` now requires `follow_symlinks`); it fails identically on `origin/dev`, so it is not introduced by this PR.
- `ruff check` and `ruff format --check` both pass on the two changed files.

---

### `custom_components/mail_and_packages/coordinator.py` — `_sum_transit_counts()`

* **Root cause correctly identified.** Since #1399, every `*_packages` sensor is a computed rollup of `_delivering + _delivered` (`GenericShipper._compute_package_totals`), with no IMAP search of its own. The old fallback `if key.endswith(("_delivering", "_packages"))` therefore counted a delivered package as in transit: delivery drops `_delivering` to 0 (which the `value <= 0` guard skips), but `_packages` stays at 1, so the rollup reports one package as both in transit and delivered until midnight reset.
* **Fix is minimal and semantically correct.** The branch now reads only `if key.endswith("_delivering")`, removing the `_packages` case entirely. This matches the documented invariant of `mail_packages_in_transit` ("total count of all packages across all carriers currently out for delivery today") in `docs/architecture.md`.
* **Invariants preserved.** Exceptions are still added for all shippers (separate `if key.endswith("_exception")` block, unaffected). `amazon_packages` keeps its special-case handling — `AmazonShipper` has its own `process_batch` and does not go through `_compute_package_totals`, so it still means "total arriving," not a rollup. `shippers_counted` still prevents `amazon_delivering` being double-added on top of `amazon_packages`.
* **Safety of dropping `_packages`:** Verified against `GenericShipper._compute_package_totals` — every `*_packages` value is always `delivering + delivered`, so it can never represent packages that are only delivered. Dropping the branch never loses a legitimately in-transit package, because any shipper with an in-transit package will also have a positive `*_delivering` value. There is no shipper that reports `_packages` without `_delivering`.

### `tests/test_coordinator.py` — regression + retargeting

* **New regression test `test_aggregate_transit_excludes_delivered_packages`** walks a single FedEx package through both states exactly as the bug report describes: out-for-delivery yields `zpackages_transit == 1`, and after delivery (with `_delivering: 0`, `_packages: 1`) yields `zpackages_transit == 0` while `zpackages_delivered == 1`. This is a precise, deterministic regression guard for the reported scenario.
* **`test_aggregate_package_counts` retargeted** to the new invariant: fedex gains `fedex_delivering: 1`, expected transit goes 13 → 11 (fedex_packages no longer counted), delivering 5 → 6. The inline arithmetic comments are clear.
* **`test_sum_transit_counts_packages_only` → `test_sum_transit_counts_ignores_packages_rollup`** correctly retargeted: the packages-only shipper no longer exists after #1399, so it now exercises the rollup invariant (`hermes_packages` rollup of 1 delivering + 1 delivered contributes only the 1 delivering).
* **Redundant `@pytest.mark.asyncio`** on `test_aggregate_transit_excludes_delivered_packages` — harmless (config sets `asyncio_mode = "auto"`), but the other tests in the file correctly omit it. Minor, not worth a change request.

**Local test results:** all 52 tests in `tests/test_coordinator.py` pass; the 4 directly affected tests pass.

---

* **No existing reviews or review comments** on PR #1400 (only a Codecov bot comment). This is a fresh review.
* **Code-only review** per operator note — no UI/frontend considerations.
* **Pre-existing failure note:** `tests/test_camera.py::test_generic_camera_rebuilds_gif_when_missing` fails locally due to anyio 4.x (`anyio.Path().exists()` requires `follow_symlinks`); it fails identically on `origin/dev` and is unrelated to this PR. The PR's CI uses a pinned test stack that passes cleanly.

firstof9:
Thanks @johnbr

PR #1401, opened 2026-09-02
fix(image): apply EXIF orientation before writing GIF frames

johnbr:
## The bug

Delivery photos are rendered 90 degrees out of true whenever the source carries an EXIF `Orientation` tag.

`generate_delivery_gif()` already tries to handle this:

```python
img = ImageOps.exif_transpose(img)  # auto-rotates according to EXIF
```

but it never sees a file with EXIF on it. Every caller runs `resize_images()` first, and `resize_images()` saves each frame as **GIF**, which carries no EXIF — so the tag is gone one step before the correction runs, and the transpose is a no-op on every real code path.

## Observed

A FedEx delivery photo off a Zebra TC78 handheld, stored 240x320 with `Orientation = 6`:

- **Email client** (honours the tag): 320x240 landscape, upright.
- **Camera entity**: 240x320, lying on its side.

It costs most of the frame too — padded as portrait it is letterboxed into the 800x600 canvas instead of filling it.

Courier handhelds routinely store rotated pixels plus an Orientation tag rather than rotating the pixels, so this is the common case for delivery photos rather than an edge case.

## The fix

Move the transpose into `resize_images()`, at the last point where the tag still exists.

That covers all four callers — the generic delivery camera plus the Amazon, USPS and Post DE shippers, which all pass originals through `resize_images()` before `generate_delivery_gif()`.

`ImageOps.exif_transpose` is a no-op for images with no Orientation tag, so images that were already correct are unaffected. The existing call in `generate_delivery_gif()` is left alone — it is still correct for anything that hands it originals directly.

## Tests

New `test_resize_images_applies_exif_orientation` builds a real portrait JPEG with `Orientation = 6` and asserts the padded result is landscape. Verified to fail on `dev` and pass with the fix.

`752 passed`, coverage 99.81%, ruff clean.

### DIFF
--- a/custom_components/mail_and_packages/utils/image.py
+++ b/custom_components/mail_and_packages/utils/image.py
@@ -188,6 +188,14 @@ def resize_images(images: list, width: int, height: int) -> list:
             img_path = Path(image_path)
             with img_path.open("rb") as fd_img:
                 img = Image.open(fd_img)
+                # Bake in the EXIF orientation before anything else. Delivery
+                # photos come straight off a courier's handheld and are very
+                # often stored rotated with an Orientation tag, and the frames
+                # written below are GIF, which cannot carry EXIF at all — so
+                # this is the last point at which the tag still exists. Doing
+                # it any later (generate_delivery_gif also calls
+                # exif_transpose) is a no-op on frames whose EXIF is gone.
+                img = ImageOps.exif_transpose(img)
                 img.thumbnail((width, height), resample=Image.Resampling.LANCZOS)
                 img = ImageOps.pad(
                     img,
--- a/tests/utils/test_image.py
+++ b/tests/utils/test_image.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from unittest.mock import MagicMock, patch
 
 import pytest
+from PIL import Image as PILImage
 
 from custom_components.mail_and_packages.const import (
     CONF_STORAGE,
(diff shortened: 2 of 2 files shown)


secondof9 (approved):
## 📋 Review Summary

> [!TIP]
> **Change Type:** 🐛 Bug Fix
> **Review Effort:** 🟢 Low
> **Core Impact:** Applies `ImageOps.exif_transpose()` inside `resize_images()` so rotated delivery photos are uprighted before their EXIF is discarded when writing GIF frames (which cannot carry EXIF).

---

| :--- | :--- | :--- |
| CI (`pytest`) | ✅ PASSED | Full suite green locally (751 passed); all 5 CI workflows completed successfully |
| Validation | ✅ PASSED | Home Assistant schema/validation clean |
| Pre-commit | ✅ PASSED | ruff/black/isort/prettier all pass |
| Lint PR | ✅ PASSED | No lint regressions |
| Autolabeler | ⏩ SKIPPED | Not applicable |

> [!NOTE]
> **CI Pipeline Clear:** All 5 GitHub Actions workflows completed successfully on head SHA `f310ad8e` (Pre-commit, Lint PR, Validation, CI). No in-progress or queued checks require waiting.

**Independent local verification:** Set up an isolated `.venv313` (Python 3.13.12, matching the repo's `Requires-Python >=3.12` constraint) and ran `pytest tests/utils/test_image.py` — all 45 tests pass, including the new `test_resize_images_applies_exif_orientation`. The full `tests/` suite passes with 99.81% coverage. The only failing test in the full run (`test_camera.py::test_generic_camera_rebuilds_gif_when_missing`) is **pre-existing** — it fails identically on `origin/dev` and is untouched by this PR. ruff is clean on both changed files.

---

### `custom_components/mail_and_packages/utils/image.py`

* **Correct root-cause fix at the right layer.** The bug is that `resize_images()` writes delivery photos to `.gif` frames, and GIF has no EXIF channel, so any `exif_transpose()` call downstream (e.g. in `generate_delivery_gif`) is a no-op on frames whose orientation tag has already been stripped by the resize/pad pipeline. Calling `ImageOps.exif_transpose(img)` *before* `thumbnail()`/`pad()` is the last point where the tag still exists — exactly the right place. This matches the author's well-reasoned inline comment.
* **The comment is excellent and self-documenting.** It explains *why* the fix must live here and not downstream, which is the kind of "why" that prevents the same regression recurring. No change needed.
* **Minimal surface area.** One line added inside the existing `try/except (OSError, ValueError)` block. `exif_transpose` is a pure in-memory transform on the already-open `PIL.Image`; it adds no I/O and no new failure modes beyond what `Image.open` already handles. If the (rare) edge case of an orientation tag that PIL cannot apply occurred, it would raise and be caught by the existing handler — no new unhandled exception path.
* **No async/executor concerns.** This is a synchronous utility operating on a local file path within the executor context; no event-loop blocking introduced, no `hass.async_add_executor_job` needed.

### `tests/utils/test_image.py`

* **Strong regression test.** `test_resize_images_applies_exif_orientation` constructs a real portrait image with `Orientation = 6` (rotate 90°), runs it through `resize_images`, and asserts the output is landscape (via the black-bar bounding-box geometry). This directly encodes the bug's symptom and would catch any regression where orientation is dropped.
* **Real fixture, not a mock.** It uses an actual JPEG with a hand-written EXIF tag rather than mocking `ImageOps.exif_transpose`, so it exercises the real code path end-to-end.
* **Correct async marking.** `@pytest.mark.asyncio` on an `async def` test — consistent with the file's existing style.

---

## ✅ Verdict

This is a clean, minimal, and well-tested bug fix. The change addresses a real-world symptom (delivery photos appearing rotated on the camera entity), targets the correct code layer, ships with a thorough regression test that uses real fixtures, and passes CI plus independent local verification. No architectural, async, or defensive-data concerns.

**Recommendation: APPROVE.**

firstof9:
Thanks @johnbr

